Issue #3, opened 2018-03-22
History view not visible in Split view


raissanorth:
While testing, when using the CMS in split view and then decreasing the screen width, the preview appears to still be enabled - even while it is not shown. 

This causes the `History` tab to not function properly. I can only see the content when loading, but then get immediately redirected to the `Content` tab.

To reproduce:

1. Open a page in the CMS
2. Select Split mode
3. Decrease the screen width
4. Click on the History tab

I am using commit 15252f3 on PHP 5.6

robbieaverill:
Nice catch! I guess we disable the preview entirely for now

robbieaverill:
This will be included with https://github.com/silverstripe/silverstripe-versioned/issues/99

robbieaverill:
Fix at #6

Issue #1, opened 2018-03-11
Allow history viewer screens to be permalinked


robbieaverill:
Originally from: https://github.com/silverstripe/silverstripe-versioned/issues/37

Original A/C from the history viewer - list versions issue:

> Each view has a permanent URL (useful e.g. for linking from workflow emails)

Separated out into a separate issue here.

NightJar:
Status update since being reassigned to other things:

The version of react-router we're using (in [`silverstripe/admin`](https://github.com/silverstripe/silverstripe-admin)) is old.
A bit of a pain, but still probably OK kinda old.

So finding docs is a matter of hitting that tag in the repo and reading the [on-github in-repo markdown](https://github.com/ReactTraining/react-router/blob/v2.4.1/docs/).
That said, it _seems_ to be a matter of registering the routes for that particular section, complete with children therein. The given example from @unclecheese does not follow the example docs on implementation, rather uses some [central registry](https://github.com/silverstripe/silverstripe-campaign-admin/blob/fa10084a3cf87ecbb60964d10970f55696922ecb/client/src/boot/index.js#L12) type approach - presumably to manage all routes within the CMS without re-making routers all over the show. The syntax appears to be as per the [react-router docs](https://github.com/ReactTraining/react-router/blob/v2.4.1/docs/guides/RouteConfiguration.md#configuration-with-plain-routes) though.

Similar to the way SS routing works there are variables in-route that allow things like PageID, VersionID etc to be passed down. So then it's mostly just a matter of adding the `<Link>` components in the appropriate places. My next (current) step was figuring out where exactly those relevant places are - I assumed the full history viewer component (list on down to viewing a specific version), but was in the process of double checking that assumption.

robbieaverill:
Observation: react-router implementations that exist already in the SilverStripe core expect a static base URL segment e.g. `admin/historyviewer`, but our implementation is relatively dynamic in that we provide a HistoryViewerField and a default CMS implementation, so the URLs are always different. This is tricky for when we register the `path` as something static in `ReactRouteRegister`

NightJar:
How does that play with the base of the router component? I presume there's one high up in the component stack that we don't get to directly play with... but is it possible to layer them (ie. pass control to a new router at the history view level), or perhaps modify (append to) the child routes of an already registered route?

robbieaverill:
We register a route with `ReactRouteRegister`, giving it the `path` and any child route patterns.

[...]

PR #5, opened 2018-04-02
Add detail view for a version

### LINKED ISSUE #7: Unmount component with entwine onunmatch()
When leaving the History tab for a page or other DataObject we should unmount the React component, to ensure that it re-renders and displays the list view again when a user goes back to the History tab.

robbieaverill:
# High level summary of this PR:

* Adds new components and flexibility to existing components for viewing the details of a version
* Adds PHP form factory and HistoryViewerController to handle rendering a version of an object, with form factory schema
* Adds redux to the components that need it
* Adds and requires a loading state from the silverstripe/admin module
* Injects sub-components as props via Injector HOCs

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/471 for the inverted badge state
- [x] https://github.com/silverstripe/silverstripe-admin/pull/490 for the loading state
- [x] https://github.com/silverstripe/webpack-config/pull/19 for the loading state

Resolves https://github.com/silverstripe/silverstripe-versioned/issues/104 and resolves #7


raissanorth:
I have run these changes on my machine and I works great!

raissanorth:
Not related to this pull request as such, but when toggling between the content and history tab one has to refresh the page to see the latest changes. 

Due to the existing (and recorded I believe) bug, the functionality  introduced in this PR currently does not work in split view, as the history list is not displayed and the content is reloaded instead.

raissanorth:
I just noticed that when a detailed view is selected and instead of closing it properly via the provided button, one just navigates to a different part of the CMS and then reopens the History tab, the detailed view is displayed, rather than the version list. Is that behaviour intended?

robbieaverill:
>Not related to this pull request as such, but when toggling between the content and history tab one has to refresh the page to see the latest changes.

Yeah, I've noted that bug above. It's pre-existing and I'll be looking into it shortly.

>Due to the existing (and recorded I believe) bug, the functionality introduced in this PR currently does not work in split view, as the history list is not displayed and the content is reloaded instead.

Reference issue is #3 

>I just noticed that when a detailed view is selected and instead of closing it properly via the provided button, one just navigates to a different part of the CMS and then reopens the History tab, the detailed view is displayed, rather than the version list. Is that behaviour intended?

Perhaps a question for @clarkepaul.

This is happening either because it's the same instance of history viewer between requests (could be "fixed" by unmounting the component with an entwine `onunmatch()` handler) or that the redux state is shared between instances (could be "fixed" by clearing the redux store when the component is unmounted).

tractorcow:
I'm going to do a bit of review and suggestions on this PR. :)

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -167,10 +203,25 @@ HistoryViewer.defaultProps = {
...
+    handleSetCurrentVersion(id) {
+      dispatch(setCurrentVersion(id));
+    },
+  };
+}
+
 export { HistoryViewer as Component };
 
-const HistoryViewerProvider = compose(
+export default compose(
```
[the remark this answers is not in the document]
tractorcow: Maybe it's ok. We need the `connect` at the very least.

client/src/containers/HistoryViewer/HistoryViewerStateRouter.js
```diff
@@ -24,6 +26,7 @@ const historyStateRouter = (RoutedComponent) => {
       return (
         <RoutedComponent
           {...this.props}
+          store={window.ss.store}
```
tractorcow: Shouldn't we rely on the connect in `components/HistoryViewer/HistoryViewer.js` only, and not inject the `window.ss.store` global here. I'll leave a note on improving the store connection. :D

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -33,6 +39,30 @@ class HistoryViewer extends Component {
...
+   * Renders the detail form for a selected version
+   *
+   * @returns {HistoryViewerVersionDetail}
+   */
+  getVersionDetail() {
+    const { currentVersion, recordId, recordClass, handleSetCurrentVersion, store } = this.props;
+
+    const sectionConfigKey = 'SilverStripe\\VersionedAdmin\\Controllers\\HistoryViewerController';
+    const sectionConfig = store.getState().config.sections
+      .find((section) => section.name === sectionConfigKey);
```
tractorcow: Config properties should be injected via HOC. In asset-admin, for example, the boot code uses this.

```js
const sectionConfig = Config.getSection('SilverStripe\\AssetAdmin\\Controller\\AssetAdmin');

reactRouteRegister.add({
    path: sectionConfig.url,
// ...
```

Maybe injecting schemaUrl could be done by the HistoryViewerRouter HOC?

HistoryViewer.js should have a schemaURL that's injected by the HOC.

Then in your unit tests, you can just inject a hard-coded URL via props.

If you like, schemaURL could have wildcards. E.g. `:id` replaced with the recordID, `:recordClass` the record class, and so on.

Instead of extracting this from redux use `import Config from 'lib/Config';` instead. :D Much less complex.

tractorcow: If you do this, you can completely omit the `window.ss.store` from the HOC anyway. :)

client/src/containers/HistoryViewer/HistoryViewerStateRouter.js
tractorcow: Actually, can you try `import Config from 'lib/Config';` and importing the section config at the HOC level, and then just inject the schemaURL property here? You don't need to use redux / store at all then.

tractorcow: If we need `store={window.ss.store}` for connect to work then I guess leave that in, but don't access `this.props.store` directly in HistoryViewer.js.

client/src/components/HistoryViewer/HistoryViewer.js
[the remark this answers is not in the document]
tractorcow: ok sounds good. :D No changes needed for this then @robbieaverill

client/src/components/HistoryViewer/HistoryViewer.js
unclecheese: We shouldn't be passing the store as a prop. Instead, use the `connect` HOC to provide the slice of config you need from the store as a prop, as @tractorcow has articulated.

The only place `store` should be passed as a prop is on the top-level component that creates the redux context. It can be accessed directly (e.g. `getState()`) but only in action creators.

client/src/containers/HistoryViewer/HistoryViewerStateRouter.js
unclecheese: FYI we can't put the store in `Injector` or expose it via `expose-loader` very easily as far as I can tell. The store is lazily build in an `Injector.ready()` callback, which means it can't really be a singleton. I think a global store object is the only way to share it at this point.

client/src/components/HistoryViewer/HistoryViewer.js
robbieaverill: Ok I've updated this to have a HistoryViewerConfig HoC which gets the store config and provides it as a prop to HistoryViewer, and the `schemaUrl` is now an injected prop too with placeholders for the variables.

client/src/containers/HistoryViewer/HistoryViewerStateRouter.js
robbieaverill: I've updated this so that it has `<Provider store={window.ss.store}><HistoryViewer ...props />...` in render() - so far as I can tell at this stage we must use `window.ss.store`, but I've removed the other direct use of the redux store so it's confined to this HoC only now.

robbieaverill:
@unclecheese I've just done a little bit of research into removing the local state for the current page in HistoryViewerStateRouter and putting it into Redux.

Currently [Griddle (used for pagination) doesn't play nicely with Redux stores](https://github.com/GriddleGriddle/Griddle/issues/626), so it might be easier to keep this in local state for now.

I've compared this against the gallery list view pagination in asset-admin, which behaves the same way. It has routing actions dispatches on page change, but looks like the current page is a prop coming from the GraphQL query only. The fast that the URL routing defines the page number is probably replacing the equivalent of my using local state.

Are you happy for me to leave it the way it is until we've done #1 ?

clarkepaul:
@robbieaverill I think that when you go the history tab it does a refresh of the view and display all versions, as apposed to a older version in detail view.

robbieaverill:
@clarkepaul cool, I've logged that here: https://github.com/silverstripe/silverstripe-versioned-admin/issues/7

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -22,27 +30,79 @@ class HistoryViewerVersion extends Component {
     return `${member.FirstName || ''} ${member.Surname || ''}`;
   }
 
+  /**
+   * Return a "clear" button to close the version, for example when used in a "detail view"
+   * context. This is shown when this version is "active", displayed with a blue background.
+   *
+   * @returns {FormAction|null}
+   */
+  getClearButton() {
```
unclecheese: `renderClearButton()`

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -1,14 +1,24 @@
...
+import HistoryViewerVersion from './HistoryViewerVersion';
 import { versionType } from 'types/versionType';
 
 class HistoryViewerVersionList extends PureComponent {
+  /**
+   * Return a string of HTML class names for the table element
+   *
+   * @returns {string}
+   */
+  getClassNames() {
```
unclecheese: We should be consistent here and use the `classnames` library.

`return classnames({ table: true, extraClass});`

client/src/components/HistoryViewer/Loading.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+/**
+ * Renders a SilverStripe loading animation
+ */
+const Loading = () => (
+  <div className="flexbox-area-grow">
```
unclecheese: I'm really surprised this isn't in our library already. Can you have a poke around and see if it seems reasonable to roll into `silverstripe/admin`? Seems like a reasonable add to me.

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -22,27 +30,79 @@ class HistoryViewerVersion extends Component {
...
         <td>{this.getAuthor()}</td>
+        { this.getClearButton() }
       </tr>
     );
   }
 }
 
 HistoryViewerVersion.propTypes = {
+  isActive: React.PropTypes.bool,
+  handleSetCurrentVersion: React.PropTypes.func,
```
unclecheese: This should be an `on` prop. Something like `onSelect` I think would be sufficient.

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -0,0 +1,61 @@
...
+  render() {
+    const { handleSetCurrentVersion, schemaUrl, version } = this.props;
+    const { loading } = this.state;
+
+    return (
+      <div className="history-viewer">
+        <HistoryViewerVersionList
+          extraClass="history-viewer__table--current"
+          versions={[version]}
+          handleSetCurrentVersion={handleSetCurrentVersion}
```
unclecheese: I don't know if we need to be passing this down the tree through several components. The advantage of redux is that you can inject this at the point in the tree where the concern is. I'd prefer to see it at the `HistoryListVersion` level. Just add a container export to that file.

client/src/legacy/HistoryViewer/HistoryViewerEntwine.js
```diff
@@ -20,6 +20,7 @@ jQuery.entwine('ss', ($) => {
       ReactDOM.render(
```
unclecheese: The lack of an `onunmatch` event here is causing a critical bug, where each match of the selector is loading another HistoryViewer component into the stack, meaning the number of renders (and network requests to the form schema) are endlessly building. 

```js
    onunmatch() {
      ReactDOM.unmountComponentAtNode(this[0]);
    }
```

This is the current output when viewing a detail after having previously viewed two other pages:

screenshot 2018-04-17 11 42 20

client/src/state/historyviewer/HistoryViewerReducer.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {object} payload - Optional data passed with the action.
+ */
+export default function historyViewerReducer(state = initialState, { type, payload } = {}) {
+  switch (type) {
+    case HISTORY_VIEWER.SET_CURRENT_VERSION: {
+      return {
+        ...state,
+        currentVersion: payload.id,
+      };
+    }
```
unclecheese: This is good, but the store persists its state after the component is unmounted, meaning when I go from the Home history view, select version 1, then to the About Us history view, Version 1 is already selected. It also means that when I go to a page that doesn't have version 1, it throws, and I have to refresh the browser.

This is an a atypical scenario, because the history viewer is an ephemeral component in a persistent environment, so all of the garbage collection has to be imperative. In a proper SPA, we don't have to worry about this.

A `componentWillUnmount()` handler that calls a reset action in one of the top level components should suffice.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -33,6 +40,36 @@ class HistoryViewer extends Component {
...
+
+    // Insert variables into the schema URL via regex replacements
+    const schemaReplacements = {
+      ':id': recordId,
+      ':class': recordClass,
+      ':version': currentVersion,
+    };
+
+    const props = {
+      schemaUrl: schemaUrl.replace(/:id|:class|:version/g, (match) => schemaReplacements[match]),
```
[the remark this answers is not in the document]
unclecheese: This looks like a good candidate. https://github.com/sindresorhus/query-string

If you can make something work, great, but I wouldn't spend too much time wrestling with it. It's a pretty minor thing. Would be nice to have a standard way of doing this, as all components are doing it differently right now.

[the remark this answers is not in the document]
robbieaverill: If you're OK with it I will log this as a separate issue. That package looks good, and I can see merit in implementing it across the SilverStripe repos that do this kind of thing. It looks like we _may_ need to use an older version of that repository that supports some older browsers - would need to be checked before implementing.

client/src/components/HistoryViewer/Loading.js
robbieaverill: I've raised some pull requests to add this to admin and dogfood it in asset-admin and campaign-admin: https://github.com/silverstripe/silverstripe-admin/pull/490

client/src/legacy/HistoryViewer/HistoryViewerEntwine.js
robbieaverill: True! This was logged as #7 but I'll fix it up in this PR too

robbieaverill:
Ok @unclecheese I've updated everything with your suggestions.

High level notes:

* Strict type checking is not in place for component props. `PropTypes.func` covers both stubs in tests and components in the wild, so I've left it as that for now. Do we want to raise an issue somewhere like admin to determine a better way of doing this and document it?
* The loading indicator has been moved to admin - required PRs to get this green are https://github.com/silverstripe/silverstripe-admin/pull/490 and https://github.com/silverstripe/webpack-config/pull/19 (note: does introducing React components from 4.2 into webpack config affect users of 4.0??)

PR #10, opened 2018-04-30
Pulls/1.0/add behat tests

raissanorth:
Adds Behat config and the first feature that checks whether the details of a record can be viewed.
Based on 34d685f7959 of pulls/1.0/connect-version-to-preview branch.


robbieaverill:
Travis is asleep =)

PR #8, opened 2018-04-10
NEW Implement React preview component when viewing a version detail

robbieaverill:
This commit also updates the versionType descriptors, overrides the CMS edit form for the history viewer controller to remove unneeded markup and adds Bootstrap as a dependency so we can use mixins.

## Dependencies

- [x] https://github.com/silverstripe/silverstripe-cms/pull/2142
- [x] https://github.com/silverstripe/silverstripe-admin/pull/478
- [x] #5 

Resolves https://github.com/silverstripe/silverstripe-versioned/issues/99

PR #11, opened 2018-04-30
Add Behat feature for viewing a version (History Viewer)

raissanorth:
Adds Behat config and the first feature that checks whether the details of a record can be viewed.


robbieaverill:
I'm not sure why Travis isn't triggering for this PR and the previous one #11, I've removed the integration and re-connected it and that didn't help. Builds for #8 are running though - it's odd. I'd like to see this passing in Travis while running the tests too before we merge it though...

PR #13, opened 2018-05-03
NEW Add revert action to detail screen

robbieaverill:
Summary:

* Adds a toolbar
* Adds a revert form action into it
* Moves the pagination state into redux
* Adds messages to the version list, e.g. "successfully reverted"
* Removes the "state router"

Parent issue: https://github.com/silverstripe/silverstripe-versioned/issues/95


robbieaverill:
admin#505 is merged, behat test restarted

robbieaverill:
https://github.com/silverstripe/silverstripe-admin/pull/505 needs to be recreated, we lost part of it during rebasing

robbieaverill:
New PR at https://github.com/silverstripe/silverstripe-admin/pull/525

robbieaverill:
admin#525 merged, we're now blocked by https://github.com/silverstripe/silverstripe-admin/issues/526

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -109,10 +145,16 @@ class HistoryViewer extends Component {
       ':version': currentVersion,
     };
 
+    const version = this.getVersions()
+      .filter((eachVersion) => eachVersion.Version === currentVersion)[0];
+    const latestVersion = this.getLatestVersion();
```
unclecheese: Use `find` to avoid looping through the whole array.

```js
const version = this.getVersions().find(version => version.Version === currentVersion);
```

client/src/components/HistoryViewer/HistoryViewerToolbar.js
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Triggers a revert action to be performed for the current record's version
+   */
+  handleRevert() {
+    const { actions: { revertToVersion }, onAfterRevert, recordId, versionId } = this.props;
+
+    revertToVersion(recordId, versionId, 'DRAFT', 'DRAFT');
+
+    onAfterRevert(versionId);
+  }
```
unclecheese: `revertToVersion` returns a Promise, right? And the mutation returns `ID`. So couldn't this be:

```js
revertToVersion(recordId, versionId, 'DRAFT', 'DRAFT')
  .then(onAfterRevert);
```

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -110,6 +110,10 @@ HistoryViewerVersion.defaultProps = {
   version: defaultVersion,
 };
 
+function mapStateToProps() {
+  return {};
+}
+
```
unclecheese: Let's just pass this in as `null` to remove some of the noise.

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -1,9 +1,27 @@
...
+   * boolean property, since the list is also used in the detail context where we don't
+   * want to clear messages from.
+   */
+  componentWillUnmount() {
+    const { onClearMessages, shouldClearMessages } = this.props;
+
+    if (shouldClearMessages) {
+      onClearMessages();
+    }
+  }
```
unclecheese: This `shouldClearMessages` prop makes me really uneasy. You're basically hijacking a prop, which is intended to be compositional, and making it procedural. Almost like `myComponent.clearMessages()`.

I feel like there's probably a better approach. Can you help me understand what the intended effect is here? Maybe the clearing of messages should be a side-effect of router state?

tests/Behat/Context/FeatureContext.php
```diff
@@ -16,39 +16,77 @@ class FeatureContext extends SilverStripeContext
...
+
+    /**
+     * @When I click on the first version
+     */
+    public function iClickOnTheFirstVersion()
+    {
+        assertNotNull($this->getLatestVersion(), 'I should see a list of versions');
+        $this->getLatestVersion()->click();
+        // Wait for the form builder to load
+        $this->getSession()->wait(3000);
```
unclecheese: This can be done more expressively in gerkin: `I wait for 3 seconds`

tests/Behat/Context/FeatureContext.php
```diff
@@ -16,39 +16,77 @@ class FeatureContext extends SilverStripeContext
      */
     public function iShouldSeeAListOfVersions()
     {
-        $page = $this->getSession()->getPage();
-        $versionsTable = $page->find('css', '.history-viewer .table');
-        assertNotNull($versionsTable, 'I should see a list of versions');
+        $this->getVersions();
```
unclecheese: It seems like this test is really vague. It might be better to test more specific outcomes like:

```
Then I should see version 2 in the list
And I should see version 3 in the list
And I should not see version 11 in the list
```

tests/Behat/Context/FeatureContext.php
```diff
@@ -16,39 +16,77 @@ class FeatureContext extends SilverStripeContext
...
 
     /**
      * @Then I should see a list of versions in descending order
      */
     public function iShouldSeeAListOfVersionsInDescendingOrder()
     {
-        $page = $this->getSession()->getPage();
-        $versions = $page->findAll('css', '.history-viewer .table tbody tr td:nth-of-type(1)');
+        $versions = $this->getVersions();
+        $this->clickVersion(current($versions));
```
unclecheese: I'm not sure I get the click event here, if you're just looking for an ordered list?

tests/Behat/Context/FeatureContext.php
```diff
@@ -16,39 +16,77 @@ class FeatureContext extends SilverStripeContext
...
+    protected function getVersions($modifier = '')
+    {
+        // Wait for the table to be visible
+        $this->getSession()->wait(3000, 'window.jQuery(".history-viewer .table").length > 0');
+
+        $versions = $this->getSession()
+            ->getPage()
+            ->findAll('css', '.history-viewer .table tbody tr' . $modifier);
+
+        assertNotEmpty($versions, 'I see a list of versions');
```
unclecheese: Not sure an assertion belongs in a helper function like this. I'd prefer to see this just return the CSS query, and have your `iShouldSeeAListOfVersions` step do the assertions, instead of just being a public wrapper for `getVersions()`

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -51,35 +64,58 @@ class HistoryViewer extends Component {
     return edges.map((version) => version.node);
   }
 
+  /**
+   * Get the latest (highest) version number from the list available. If we are not on page
+   * zero then it's always false, because there are higher versions that we aren't aware of
+   * in this context.
```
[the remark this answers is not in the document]
robbieaverill: Yeah it does, right now you can't sort the list

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -51,35 +64,58 @@ class HistoryViewer extends Component {
...
+   *
+   * @returns {object}
+   */
+  getLatestVersion() {
+    const { page } = this.props;
+
+    if (page > 1) {
+      return false;
+    }
```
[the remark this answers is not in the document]
robbieaverill: I think this is the zero-based Griddle pagination thing again, will re-check

client/src/components/HistoryViewer/HistoryViewerVersion.js
robbieaverill: I had `{}` before and it didn't work, which is why I've moved it to a function. I'll try it again with `null` instead =)

client/src/components/HistoryViewer/HistoryViewerVersionList.js
robbieaverill: Sure - so the component should render any messages from the redux state, once. After rendering them it should clear them again. I had some trouble getting the right logic for this without causing infinite re-rendering loops in React =/

tests/Behat/Context/FeatureContext.php
robbieaverill: True, I thought we'd actually replaced this with a javascript condition to match the form builder. I didn't want to pollute the user stories with statements like this, since the human doesn't really concern themselves with it - I guess though it is true, they will have to wait for the form builder to load...

tests/Behat/Context/FeatureContext.php
robbieaverill: You do have that assertion as well

tests/Behat/Context/FeatureContext.php
robbieaverill: You're right - this has nothing to do with "seeing a list of versions in descending order"

client/src/components/HistoryViewer/HistoryViewerToolbar.js
raissanorth: Tried changing it, but it breaks the test.

unclecheese: That would be my expectation, because you've changed it from an asynchronous call to a synchronous one. A failing test shouldn't be a reason to not implement it. It should be a reason to update the test. :) 

At the very least, change the prop name from `onAfterRevert` to something else, because it's if it's being invoked asynchronously, that's not an honest name.

unclecheese: Actually, `onRevert` is a more appropriate name.

raissanorth:
@unclecheese Thanks again for your feedback and help! If you find some time, could you please have a look at the changes we made, based on your suggestions. Thanks!

robbieaverill:
Yep, looks great. Thanks for your work finishing this off!

PR #18, opened 2018-05-15
NEW Allow history viewer to be activated by modules rather than always on

robbieaverill:
Resolves  https://github.com/silverstripe/silverstripe-versioned-admin/issues/15

This allows the data record to provide context to whether the history viewer should be activated or not. Example activation point: https://github.com/dnadesign/silverstripe-elemental/pull/231

### DIFF
--- a/_config/cms.yml
+++ b/_config/cms.yml
@@ -5,7 +5,8 @@ Only:
 ---
 SilverStripe\Core\Injector\Injector:
   SilverStripe\CMS\Controllers\CMSPageHistoryController:
-    class: SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController
+    factory:
+      SilverStripe\VersionedAdmin\Controllers\HistoryControllerFactory
 
 SilverStripe\CMS\Controllers\CMSMain:
   extensions:
--- a/behat.yml
+++ b/behat.yml
@@ -21,7 +21,7 @@ default:
             - %paths.modules.versioned-admin%/tests/behat/files/
         -
           SilverStripe\Framework\Tests\Behaviour\ConfigContext:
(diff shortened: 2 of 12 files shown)


robbieaverill:
@unclecheese recently you suggested it'd be nice to make this a controller factory that resolves based on the input SiteTree object's extensions to the appropriate controller. I totally agree! @raissanorth and I have both tried to overload the `/admin/pages/history` route with a class factory - the main problem is that the constructor doesn't get provided the SiteTree object in its constructor, so you'd end up doing it in a semi stateless way and probably end up with similar code to what I've got here anyway.

The example code you sent from `AdminRootController::handleRequest()` doesn't provide any arguments when instantiating the controller:

```php
        $rules = self::rules();
        foreach ($rules as $pattern => $controller) {
            if (($arguments = $request->match($pattern, true)) !== false) {
                /** @var LeftAndMain $controllerObj */
                $controllerObj = Injector::inst()->create($controller); // <----------------------
                return $controllerObj->handleRequest($request);
            }
        }
```

Thoughts on proceeding?

unclecheese:
This was the Injector solution we ended up with:

```yaml
SilverStripe\Core\Injector\Injector:
  SilverStripe\CMS\Controllers\CMSPageHistoryController:
    factory:
      MyApp\TestModule\MyFactory
```

```php
class MyFactory implements Factory
{
    public function create($service, array $params = array())
    {
        $request = Injector::inst()->get(HTTPRequest::class);
        $id = $request->getParam('ID');
        if ($id) {
            $page = SiteTree::get()->byID($id);
            if ($page && $page instanceof MySpecialPage) {
                return new SpecialPageHistory();
            }
        }

        return new CMSPageHistoryController();
    }
}
```

PR #19, opened 2018-06-07
FIX Ensure history controller factory reads current page from Draft stage

robbieaverill:
In #13 this change had originally made a behat test fail, perhaps the test has been updated so workaround that, but this is the fix to ensure that the page can be loaded when the reading mode is Stage.Live


robbieaverill:
Self merging with @NightJar's team approval, since this change should have been part of the original PR and is important for people to actually use the history viewer

PR #23, opened 2018-06-15
FIX Lower URL priority so controller is not auto loaded over CMSPageHistoryController

robbieaverill:
We use the HistoryControllerFactory to determine this instead.

This meant that in practice the factory response was being ignored in favour of this controller anyway. Testing with this fix in place resolves the issue.

Fixes https://github.com/silverstripe/cwp/issues/110


robbieaverill:
The behat failure is because the reverting feature doesn't include the necessary config to enable history viewer everywhere. This would've been a pre-existing bug, not sure why it wasn't failing already. Fixed in the later commit.

NightJar:
I cannot merge this, but it looks good :)

robbieaverill:
Self merging with @NightJar's approval (team) since it's a low risk high value patch

PR #25, opened 2018-06-27
Install 5.x-dev recipe when testing master.

maxime-rainville:
Parent issue silverstripeltd/open-sourcerers#31


robbieaverill:
We can fix Behat later

PR #46, opened 2018-08-19
Add 'already selected' message for version chosen in compare mode

raissanorth:
Adds an 'already selected' message to the version chosen in compare mode. This is added to each element in the compare mode, but hidden by default unless the version is _the_ selected version. On hover and or focus via keyboard the message is replaced by the close button .

Chosen version in compare mode:

image

Chosen version in compare mode on hover/focus:

image


client/src/styles/bundle.scss
```diff
@@ -13,3 +13,4 @@
 @import "../components/HistoryViewer/HistoryViewerHeading";
 @import "../components/HistoryViewer/HistoryViewerVersion";
 @import "../components/HistoryViewer/HistoryViewerVersionList";
+@import "../components/HistoryViewer/HistoryViewerVersion";
```
robbieaverill: I missed this, this import already exists

Issue #47, opened 2018-08-19
Issues with compare mode on Edge and IE


raissanorth:
**Windows / Edge**

The compare mode link is not centered above the list:

![screen shot 2018-08-20 at 10 58 04 am](https://user-images.githubusercontent.com/14869519/44314083-60a85a00-a468-11e8-9ec1-1fd96f5e986e.png)

When two versions are selected and the diffs are rendered half the screen is reserved for a preview:

![screen shot 2018-08-20 at 10 58 27 am](https://user-images.githubusercontent.com/14869519/44314084-61d98700-a468-11e8-8242-95d12eb33dc9.png)

When unselected both or one of the selected versions, half the screen is still reserved for a preview:

![screen shot 2018-08-20 at 10 58 50 am](https://user-images.githubusercontent.com/14869519/44314085-630ab400-a468-11e8-931e-860dbe011417.png)

![image](https://user-images.githubusercontent.com/14869519/44314120-f93eda00-a468-11e8-8387-913b1677a2b5.png)

**Windows / Internet Explorer**

Neither hover nor focusing via keyboard are triggering the CSS to show the close button for the selected version instead of showing the 'already selected' message - see https://github.com/silverstripe/silverstripe-versioned-admin/pull/46.

raissanorth:
### Outstanding work:

- [x] Center compare mode link above the list - _addressed by_ https://github.com/silverstripe/silverstripe-versioned-admin/pull/48
- [ ] Spread version list over entire width of window, rather than reserving space for phantom preview
- [ ] Show close button on hover for IE

PR #48, opened 2018-08-20
Center alert message and conditionally show close button/selected mes…

raissanorth:
Addresses _some_ of the issues raised in https://github.com/silverstripe/silverstripe-versioned-admin/issues/47 in that it centers the alert message for IE 11 and shows the 'Already selected message' for the chosen version in compare mode on hover. 

`Focus-within` is not supported by Edge and IE so the styles targeted by the `hover` selector were not applied. The solution does not follow the DRY principle strictly, but it works. Happy to make changes as needed.

Screenshot from BrowserStack using Windows 10 and IE 11:

image


robbieaverill:
Travis reckons the dist files either haven't been built or the branch needs to be rebased. Could you take a look at that?

Also, I assume you've tested these changes on the other browsers to check for regressions there?

robbieaverill:
Also, yay for removing absolute positioning!

raissanorth:
Checked IE, Edge, Chrome, Opera and Firefox. Unfortunately had to add the absolute positioning, because otherwise the entire 'row' containing the message would float above the list; this would cause some light grey to the left and right of the message to overlay the white version list items.

PR #54, opened 2018-08-27
FIX Remove any query string arguments from the schema config URL

robbieaverill:
See https://github.com/silverstripe/silverstripe-admin/issues/616

This the fix for this module


robbieaverill:
Moving to FormBuilderLoader...

PR #66, opened 2018-10-08
MINOR Upgrade Bootstrap to 4.1.2

robbieaverill:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/8392

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "@storybook/react": "^3.2.16",
     "apollo-client": "^0.5.22",
     "babel-polyfill": "6.7.4",
-    "bootstrap": "4.0.0",
+    "bootstrap": "4.1.2",
     "classnames": "^2.2.5",
     "graphql-fragments": "^0.1.0",
     "graphql-tag": "^0.1.17",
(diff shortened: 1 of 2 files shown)


robbieaverill:
Thanks for the response. I'm not sure I agree, to be honest. This (the bootstrap change) is a minor version bump, which I've already found one change which could break user code. We also don't use any of the XSS affected code in our implementations. Secondly, we are not treating this as a SilverStripe security issue, which is why I've made the PRs public and have labelled them as "MINOR" - perhaps best to discuss with @chillu and let me know what y'all decide.

If everyone wants this as a patch (against 4.0.x, since we support that too) then we can backport, but I just want to express my concern now with doing it that way (4.0.x runs 4.0.0-beta, and we have supporting modules that rely on some of the breaking changes from 4.0.0-beta to 4.0.0 not being there - example, input group markup changed in these versions and broke our stable elemental release).

robbieaverill:
Also note that the issue is triaged as an enhancement (although it's also triaged as a patch change, which doesn't happen in semver)

PR #68, opened 2018-10-16
Archive add missing tab logic, fix file list

lukereative:
This fixes a couple of small issues in archive admin:

1. `ArchiveViewProvider` had an `isArchiveFieldEnabled` function for determining if a given archive view should be presented as a tab in 'Archives' – currently uses for 'Files' archive tab to only show if the setting for assets to be retained when deleted is set. This function wasn't actually leveraged by ArchiveAdmin so has been added.

2. There was an issue with the files gridfield showing the count for records that weren't displayed due to canView permissions, so we do that check sooner so the paginator and page count report the correct count.

### DIFF
--- a/src/ArchiveAdmin.php
+++ b/src/ArchiveAdmin.php
@@ -5,7 +5,7 @@ namespace SilverStripe\VersionedAdmin;
 use SilverStripe\Admin\ModelAdmin;
 use SilverStripe\Control\Controller;
 use SilverStripe\Core\ClassInfo;
-use SilverStripe\Core\Injector\Injectable;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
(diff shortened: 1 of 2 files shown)


src/ArchiveAdmin.php
```diff
@@ -294,12 +294,16 @@ public function getManagedModelTabs()
...
         $mainModels = $this->getVersionedModels('main', true);
         foreach ($mainModels as $class => $title) {
-            $forms->push(new ArrayData(array(
-                'Title' => $title,
-                'ClassName' => $class,
-                'Link' => $this->Link($this->sanitiseClassName($class)),
-                'LinkOrCurrent' => ($class === $this->modelClass) ? 'current' : 'link'
-            )));
+            if (ClassInfo::hasMethod(Injector::inst()->get($class), 'isArchiveFieldEnabled')
+                && Injector::inst()->get($class)->isArchiveFieldEnabled()
```
robbieaverill: Side note: the duplicated use of Injector::inst()->get() could be avoided

src/ArchiveAdmin.php
```diff
@@ -294,12 +294,16 @@ public function getManagedModelTabs()
...
+            ) {
+                $forms->push(new ArrayData([
```
robbieaverill: Please use `ArrayData::create()` (I realise that's not a new change)

src/ArchiveAdmin.php
lukereative: It's now avoided

Issue #67, opened 2018-10-10
`CMSPageHistoryViewerController` should be default `CMSPageHistoryController` in SS 4.3


maxime-rainville:
`CMSPageHistoryViewerController` has achieved feature parity with `CMSPageHistoryController`.

It should be the default controller for displaying Page History information when versioned-history is installed. An entry to the 4.3 change log should explain how to revert to the old controller.

Do we want to mark `CMSPageHistoryController` as deprecated in SS5 onwards?

# Pull requests
* [x] https://github.com/dnadesign/silverstripe-elemental/pull/463
* [x] https://github.com/silverstripe/silverstripe-versioned-admin/pull/71
* [x] https://github.com/silverstripe/silverstripe-framework/pull/8508
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2298
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2300

chillu:
@NightJar @robbieaverill Is there any reason this can't be done for 4.3? The goal of these shared core UI components should be to roll them out in all use cases, so we can't call it "done" just because Elemental uses it. Next time, if your team doesn't have the capacity to get it into that state (so planning for a partial roll out), please ensure that there's a follow up ticket which can be triaged and prioritised between teams. /cc @brynwhyman

chillu:
> Do we want to mark CMSPageHistoryController as deprecated in SS5 onwards?

There's going to be so many UI controller changes in SS 5.x (ideally they all go away) that I think it's not a high priority. We should clearly spell out the relationship between the old and new controllers though (X replaces Y in both PHPDocs)

robbieaverill:
> Is there any reason this can't be done for 4.3?

Nope. We refrained from shipping it by default with 4.2 because it didn't have compare mode. Now that it does it can be enabled for all pages.

We need to remove most of the logic from `HistoryControllerFactory::create()`, leaving the functionality for `isEnabled()` (with its extension) point to opt for the old controller still, because semver.

maxime-rainville:
Fixing this issue will probably fix https://github.com/silverstripe/silverstripe-cms/issues/2277#issuecomment-428815482 as well.

chillu:
[...]

PR #71, opened 2018-10-22
API Deprecate HistoryControllerFactory and default to CMSPageHistoryV…

maxime-rainville:
Fix https://github.com/silverstripe/silverstripe-versioned-admin/issues/67 and https://github.com/silverstripe/silverstripe-cms/issues/2277

### DIFF
--- a/_config/cms.yml
+++ b/_config/cms.yml
@@ -5,8 +5,7 @@ Only:
 ---
 SilverStripe\Core\Injector\Injector:
   SilverStripe\CMS\Controllers\CMSPageHistoryController:
-    factory:
-      SilverStripe\VersionedAdmin\Controllers\HistoryControllerFactory
+    class: SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController
 
 SilverStripe\CMS\Controllers\CMSMain:
   extensions:
--- a/src/Controllers/HistoryControllerFactory.php
+++ b/src/Controllers/HistoryControllerFactory.php
@@ -14,7 +14,7 @@ use SilverStripe\Versioned\Versioned;
  * The history controller factory decides which CMS history controller to use, out of the default from the
  * silverstripe/cms module or the history viewer controller from this module, depending on the current page type
  *
- * @deprecated 1.0.0:2.0.0
+ * @deprecated 1.1.0:2.0.0
  */
 class HistoryControllerFactory implements Factory
 {
@@ -36,6 +36,7 @@ class HistoryControllerFactory implements Factory
                 );
 
                 if ($page && !$this->isEnabled($page)) {
+                    // Injector is not used to prevent an infinite loop
                     return new CMSPageHistoryController();
                 }
             }
@@ -54,7 +55,7 @@ class HistoryControllerFactory implements Factory
      */
     public function isEnabled(SiteTree $record)
     {
-        $enabledResults = array_filter($this->extend('updateIsEnabled', $record));
-        return (empty($enabledResults) || min($enabledResults) === false);
+        $enabledResults = $this->extend('updateIsEnabled', $record);
+        return (empty($enabledResults) || min($enabledResults) !== false);
     }
 }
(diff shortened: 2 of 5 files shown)


src/Controllers/HistoryControllerFactory.php
```diff
@@ -21,31 +23,30 @@ class HistoryControllerFactory implements Factory
...
+
+                if ($page && !$this->isEnabled($page)) {
+                    return new CMSPageHistoryController();
+                }
             }
         }
 
         // Injector is not used to prevent an infinite loop
-        return new CMSPageHistoryController();
+        return Injector::inst()->create(CMSPageHistoryViewerController::class);
```
ScopeyNZ: This doesn't cause an infinite loop any more?

maxime-rainville: It would cause an infinite loop if you called it on `CMSPageHistoryController`, but not on `CMSPageHistoryViewerController`.

ScopeyNZ: Yep that makes sense

_config/cms.yml
```diff
@@ -5,8 +5,7 @@ Only:
 ---
 SilverStripe\Core\Injector\Injector:
   SilverStripe\CMS\Controllers\CMSPageHistoryController:
-    factory:
-      SilverStripe\VersionedAdmin\Controllers\HistoryControllerFactory
+    class: SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController
```
robbieaverill: This will break any implementations that use the existing API to opt out of elemental using this new feature (plus any third party modules plus user code that does the same). I think it’s pretty low risk and worth doing anyway but figure it’s worth noting.

src/Controllers/HistoryControllerFactory.php
```diff
@@ -13,6 +13,8 @@
 /**
  * The history controller factory decides which CMS history controller to use, out of the default from the
  * silverstripe/cms module or the history viewer controller from this module, depending on the current page type
+ *
+ * @deprecated 1.0.0:2.0.0
```
robbieaverill: 1.1.0*

src/Controllers/HistoryControllerFactory.php
```diff
@@ -21,31 +23,30 @@ class HistoryControllerFactory implements Factory
...
+            if ($id && is_numeric($id)) {
+                // Ensure we read from the draft stage at this position
+                $page = Versioned::get_one_by_stage(
+                    SiteTree::class,
+                    Versioned::DRAFT,
+                    sprintf('"SiteTree"."ID" = \'%d\'', $id)
+                );
+
+                if ($page && !$this->isEnabled($page)) {
+                    return new CMSPageHistoryController();
```
robbieaverill: Can you retain the comment about not using injector here?

src/Controllers/HistoryControllerFactory.php
```diff
@@ -54,6 +55,6 @@ public function create($service, array $params = array())
     public function isEnabled(SiteTree $record)
     {
         $enabledResults = array_filter($this->extend('updateIsEnabled', $record));
-        return (!empty($enabledResults) && max($enabledResults) === true);
+        return (empty($enabledResults) || min($enabledResults) === false);
```
robbieaverill: Doesn’t this logic now evaluate to “not enabled”?

src/Controllers/HistoryControllerFactory.php
maxime-rainville: Fixed.

src/Controllers/HistoryControllerFactory.php
maxime-rainville: Done.

_config/cms.yml
maxime-rainville: Yeah ... that was my thought as well.

If you care enough about what history viewer you are using to customise it, you'll probably want to update your customisation based on the updated availability `CMSPageHistoryViewerController`.

PR #72, opened 2018-10-24
Various Archive Admin improvements/fixes

lukereative:
- Improve archive list query speed
- Only show restore action on files that have their binary saved
- Use version record info for archive date and author


lukereative:
The new list query is based on https://github.com/silverstripe/silverstripe-versioned/blob/1/src/GraphQL/Resolvers/ApplyVersionFilters.php

Issue #87, opened 2018-11-26
"is preview enabled" logic should be extensible


robbieaverill:
**Affected version: all**

The logic that determines whether a DataObject is previewable is set to return `$obj instanceof CMSPreviewable`. There are cases where you may want to disable previews for objects that implement this interface though, e.g. for pages.

Devs can modify the schema data of the implementation to achieve this, but it should be easy via an extension point. Marking as a bug because it should've been designed this way originally.

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-versioned-admin/pull/88

robbieaverill:
PR at #88

robbieaverill:
Yep!

PR #88, opened 2018-11-26
FIX HistoryViewerField::getPreviewEnabled() can now be adjusted with extensions

### LINKED ISSUE #89: Padding is out and missing scrollable container when an object is not previewable
**Version: SS 4.3.x-dev**

If you bring in #88 and disable previews for a page, you see this on the list view:

image

The container also isn't scrollable:

image

robbieaverill:
* Allows getPreviewEnabled to be extensible (marking as a bugfix because we should've designed it this way originally)
* Changes the logic for whether to make the HistoryViewer container padded and scrollable from "do, if it's previewable" to "do, unless it's in a GridField." I've tested this with previewable Pages, non-previewable Pages and non-previewable Elements inside a GridFieldDetailForm and it looks good now.
* Ensures that the test stub is written - the Related Pages logic that changed recently in SiteTree fails when the model is unsaved - will raise a separate issue for this -> https://github.com/silverstripe/silverstripe-cms/issues/2343

Resolves https://github.com/silverstripe/silverstripe-versioned-admin/issues/87 and resolves #89


blueo:
This appears to cause a bit of a visual regression. If I use the extension point to return false I get the following version list display (note the version numbers aren't fully visible).

screen shot 2018-11-27 at 12 00 29 pm

robbieaverill:
Good times. We have some modifiers to adjust the UI padding, will find out tomorrow which you need to apply

robbieaverill:
@blueo this PR now has the padding/scrollable fixes in it too

robbieaverill:
The behat test failure is valid, and is showing https://github.com/silverstripe/silverstripe-cms/issues/2341 in action

ScopeyNZ:
I reran the tests for you now that the other PR has been merged and they're all green.

PR #105, opened 2019-03-14
BUG Bump bootstrap and merge js dependency

maxime-rainville:
Just update the JS dependencies


maxime-rainville:
I don't think we directly use `merge`. It only gets installed because some other dependency as a dependency on it. I think we should only explicitly require a dependency if we make direct use of it.

robbieaverill:
Yeah I agree. Maybe in future we can reference something like "update Bootstrap" in commit messages, assuming that its own dependencies are also updated

chillu:
Travis failed

PR #108, opened 2019-04-17
FIX Use Firefox compliant polyfill for ResizeAware

robbieaverill:
See https://github.com/silverstripe/silverstripe-admin/pull/818


robbieaverill:
@maxime-rainville done, thanks

Issue #102, opened 2019-03-05
ArchiveAdmin requires CMS JS/CSS


mlewis-everley:
As in the title, `ArchiveAdmin` requires CMS dependencies, but doesn't seem to need them (I commented them out and it still seems to work fine).

The code is located at: https://github.com/silverstripe/silverstripe-versioned-admin/blob/master/src/ArchiveAdmin.php#L53

If the css/js are needed, is it worth adding those requirements inside a check to see if the CSM module is installed?

I can probably add this as a PR if that helps?

ScopeyNZ:
🤔 

Please, open a PR and we can debate it there. #29 introduced this and it's a React app so all of its requirements should be within its own bundle.

robbieaverill:
@mlewis-everley still keen to submit a PR?

mlewis-everley:
I can do, I will leave this tab open so I remember to do it!

mlewis-everley:
@robbieaverill PR made

robbieaverill:
Merged, thanks

mlewis-everley:
Cheers @robbieaverill , do you need me to create a PR for 1.2 branch, or is that in hand?

robbieaverill:
Nope, it'll get merged up eventually =)

PR #123, opened 2019-07-18
Remove CMS requirement from ArchiveAdmin

mlewis-everley:
See issue #102 ArchiveAdmin appears to require CMS JS/CSS but apparently doesn't need them.

I have done this on branch 1.1 as this appears to be the earliest point `ArchiveAdmin` exists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: php
 
+dist: trusty
+
 before_install:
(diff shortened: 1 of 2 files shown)


robbieaverill:
If you add `dist: trusty` to the .travis.yml file after the first line, it should fix the builds

PR #124, opened 2019-07-22
FIX Use a more reliable method of including client bundles

ScopeyNZ:
The extension that provided this config was only applying to CMSMain which would not help history viewers that weren't loading outside of a CMS page context (eg. ModelAdmin). This PR makes the HistoryViewer JS apply more globally using a method similar to other modules that provide fields.

### DIFF
--- a/src/Extensions/CMSMainExtension.php
+++ b/src/Extensions/CMSMainExtension.php
@@ -3,7 +3,6 @@
 namespace SilverStripe\VersionedAdmin\Extensions;
 
 use SilverStripe\Core\Extension;
-use SilverStripe\View\Requirements;
 
 /**
  * @deprecated 1.2 This extension no longer serves a purpose. Write a standalone extensions instead.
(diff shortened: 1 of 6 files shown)


src/Extensions/CMSMainExtension.php
```diff
@@ -5,15 +5,12 @@
 use SilverStripe\Core\Extension;
 use SilverStripe\View\Requirements;
 
+/**
+ * @deprecated 1.2 This extension no longer serves a purpose. Write a standalone extensions instead.
+ */
 class CMSMainExtension extends Extension
```
ScopeyNZ: I left this extension because it'd be breaking to remove it - assuming someone might've `MyExtension extends CMSMainExtension`.

src/Extensions/CMSMainExtension.php
```diff
@@ -3,17 +3,13 @@
 namespace SilverStripe\VersionedAdmin\Extensions;
 
 use SilverStripe\Core\Extension;
-use SilverStripe\View\Requirements;
 
+/**
+ * @deprecated 1.2 This extension no longer serves a purpose. Write a standalone extensions instead.
```
robbieaverill:
```suggestion
 * @deprecated 1.2.0 This extension no longer serves a purpose. Write a standalone extensions instead.
```

PR #126, opened 2019-08-13
Add overflow: auto to .history-viewer class

blueo:
prevents long edit forms from breaking the CMS layout


client/src/components/HistoryViewer/HistoryViewer.scss
```diff
@@ -1,4 +1,6 @@
 .history-viewer {
+  overflow: auto;
```
robbieaverill: We usually add `.panel--scrollable` to achieve this. Would you mind trying that out instead? See here:

https://github.com/silverstripe/silverstripe-versioned-admin/blob/71e1ae3fdb29e8b8fa44204fdcd1a99fab8420be/client/src/components/HistoryViewer/HistoryViewer.js#L76-L89

blueo: ah though that might be the case but hesitated as i wasn't sure it was a 'panel'. Will give it a go

blueo: works well :)

PR #130, opened 2019-10-09
FIX Gracefully handle lack of versions in HistoryViewer

dnsl48:
Initially reported as a trigger for https://github.com/dnadesign/silverstripe-elemental/issues/736

The `actions` structure isn't defined by the proptypes, so we cannot rely on `versions` being presented there.


dnsl48:
haha, I completely messed up this PR, sorry :D 
yeah, it's supposed to land to `1.1` and I realized that I forgot to build the dist after I created it. Thanks for reminding too :)

dnsl48:
the tests are green. merging

PR #135, opened 2019-11-21
Changed "revert" button title when revert is possible.

manja:
If revert wasn't already in process, button was always showing 'Unavailable for the current version' message - both for the latest version (which clearly one can't revert to) and all other versions - where revert is possible.

Added a check should it display or not.

### DIFF
--- a/client/src/components/HistoryViewer/HistoryViewerToolbar.js
+++ b/client/src/components/HistoryViewer/HistoryViewerToolbar.js
@@ -36,7 +36,7 @@ class HistoryViewerToolbar extends Component {
 
     const revertButtonTitle = isReverting
       ? i18n._t('HistoryViewerToolbar.REVERT_IN_PROGRESS', 'Revert in progress...')
-      : i18n._t('HistoryViewerToolbar.REVERT_UNAVAILABLE', 'Unavailable for the current version');
+      : (isLatestVersion ? i18n._t('HistoryViewerToolbar.REVERT_UNAVAILABLE', 'Unavailable for the current version') : '');
 
     return (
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request).

ScopeyNZ:
As this is a bugfix, you should target your fix to a patch branch. For example, if you're running `versioned-admin@1.2.3` then you probably want to target the `1.2` branch. This should help with the failing builds (maybe).

manja:
But build errors are PHP related, like:

1) SilverStripe\VersionedAdmin\Tests\Controllers\HistoryControllerFactoryTest::testCreateController
Error: Class 'SilverStripe\CMS\Controllers\CMSPageHistoryController' not found

which has nothing to do with the nature of this PR since changes are made in JS/React part.

Anyway, should I close this PR and make a new one in 1.2 branch or do something else?

ScopeyNZ:
Yeah I'm saying it'll probably help with the builds because the master branch might be failing right now anyway. `master` is a branch for the next major version, which is years away.

robbieaverill:
I suggest that one of the core team merge this by hand into minor release branch

bergice:
Resolved through https://github.com/silverstripe/silverstripe-versioned-admin/pull/145

PR #141, opened 2019-12-17
Update BlockArchiveExtension.php

### LINKED ISSUE #118: Error: Uncaught BadMethodCallException when viewing block archives when a descendant of ElementList has been archived
Relevant package versions:
SilverStripe Versioned 1.4.1
SilverStripe Versioned Admin 1.2.1
SilverStripe Framework 4.4.1
Elemental 4.1
Elemental List 1.2

On a site I am developing, when viewing the block archive page (/admin/archive/DNADesign-Elemental-Models-BaseElement), the page breaks with an error:

```
[Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'Breadcrumbs' does not exist on '[...descendant of ElementList...]'

Trace
SilverStripe\View\ViewableData->__call(Breadcrumbs, Array) 
BlockArchiveExtension.php:51
```

I believe the block's parent is being assumed to be a page, rather than another element, and thus the `Breadcrumbs` method doesn't exist on that element when it's an ElementList instead.

If I have time I might try to submit a PR for this but not sure if I have time, I'd appreciate someone else looking into this.

jinjie:
Return Breadcrumbs only if the method exists in the parent.

Fix #118

### DIFF
--- a/src/Extensions/BlockArchiveExtension.php
+++ b/src/Extensions/BlockArchiveExtension.php
@@ -48,7 +48,7 @@ class BlockArchiveExtension extends DataExtension implements ArchiveViewProvider
             'Breadcrumbs' => function ($val, $item) {
                 $parent = $item->Page;
 
-                return $parent ? $parent->Breadcrumbs() : null;
+                return (($parent && method_exists($parent, 'Breadcrumbs')) ? $parent->Breadcrumbs() : null);
             },
             'Versions.first.LastEdited' => function ($val, $item) {
                 return DBDatetime::create_field('Datetime', $val)->Ago();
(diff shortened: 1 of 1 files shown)


src/Extensions/BlockArchiveExtension.php
```diff
@@ -48,7 +48,7 @@ public function getArchiveField()
             'Breadcrumbs' => function ($val, $item) {
                 $parent = $item->Page;
 
-                return $parent ? $parent->Breadcrumbs() : null;
+                return (($parent && method_exists($parent, 'Breadcrumbs')) ? $parent->Breadcrumbs() : null);
```
robbieaverill:
```suggestion
                return ($parent && $parent->hasMethod('Breadcrumbs')) ? $parent->Breadcrumbs() : null;
```

This would allow extensions to provide the Breadcrumbs method as well as the base class

emteknetnz:
Closed in favour of https://github.com/silverstripe/silverstripe-versioned-admin/pull/168 since this one is on the wrong branch and was quicker to just redo

I've put you both down as co-authors in the commit message

PR #160, opened 2020-07-03
Show the type of save that was performed for the record instead of just showing either Saved/Published.

bergice:
See https://github.com/silverstripe/silverstripe-cms/issues/2448


client/src/components/HistoryViewer/HistoryViewerVersionState.js
```diff
@@ -18,16 +18,27 @@ class HistoryViewerVersionState extends Component {
...
+
     if (version.Published) {
+      if (version.Deleted) {
+        if (version.Draft) {
+          return i18n._t('HistoryViewer.Archived', 'Archived');
+        }
+        return i18n._t('HistoryViewer.Unpublished', 'Unpublished');
+      }
       return i18n._t('HistoryViewer.Published', 'Published');
     }
```
robbieaverill, line 40: Do you need to nest these three if statements? IMO it would be cleaner if you didn't

client/src/components/HistoryViewer/HistoryViewerVersionState.js
```diff
@@ -18,16 +18,27 @@ class HistoryViewerVersionState extends Component {
...
 
   /**
-   * When the record is published, return "Published", else return "Saved"
+   * Return the type of action that was performed when the record was saved
    *
-   * @returns {string}
+   * @returns {string} Returns either Saved, Created, Archived, Unpublished or Published
    */
-  getPublishedState() {
+  getVersionStage() {
```
robbieaverill, line 25: This is a breaking change, please revert

client/src/components/HistoryViewer/HistoryViewerVersionState.js
bergice: I tried the other way and personally I found this way to make more sense.

robbieaverill: Ok cool, I guess if you read the logic verbatim then not nesting it wouldn't be that pretty

ScopeyNZ: I agree with Robbie here, and not nesting would be consistent with the line above where "Created" is returned.

emteknetnz:
Behat failure is likely to be because it needs corresponding versioned pr merged first

PR #165, opened 2020-08-14
FIX Reorder model tabs

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/207

You can get the Files tab to easily show by return true from `FileArchiveExtension::isArchiveFieldEnabled()`

(most installations will probably have this set to false and won't show the Files tab)

Default tab is selected

Before
image

After
image

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -7,6 +7,6 @@ en:
     COLUMN_TYPE: Type
     SELECT_EMPTY: Select…
     SELECT_TYPE: 'Select a content type'
-    TAB_OTHERS: Others
+    TAB_OTHERS: Other
   SilverStripe\VersionedAdmin\Controllers\HistoryViewerController:
     ErrorItemViewPermissionDenied: 'You don''t have the necessary permissions to view {ObjectTitle}'
--- a/src/ArchiveAdmin.php
+++ b/src/ArchiveAdmin.php
@@ -45,6 +45,13 @@ class ArchiveAdmin extends ModelAdmin
     protected function init()
     {
         parent::init();
+
+        // Set the default model class to SiteTree, as long as silverstripe/cms is installed
+        // This is done otherwise File will be the default set in ModelAdmin::init() which is basically random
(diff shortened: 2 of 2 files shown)


emteknetnz:
image

lang/en.yml
```diff
@@ -7,6 +7,6 @@ en:
     COLUMN_TYPE: Type
     SELECT_EMPTY: Select…
     SELECT_TYPE: 'Select a content type'
-    TAB_OTHERS: Others
+    TAB_OTHERS: Other
```
robbieaverill, line 10: Can you create a new key? `TAB_OTHER` seems like a nice name

src/ArchiveAdmin.php
```diff
@@ -324,6 +331,13 @@ public function getManagedModelTabs()
         $forms = ArrayList::create();
         $mainModels = $this->getVersionedModels('main', true);
 
+        // Display order should be Pages > Blocks > Files > Other
+        // Pages, Blocks, Files are treated specially and have extensions defined in _config/archive-admin.yml
```
robbieaverill, line 335: Where do user defined tabs come into this? (Aside from "other")

lang/en.yml
emteknetnz: But we're changing an existing label to something else, seems like not updating the key is the correct thing.  Also it'll mean that any other languages will loose their translation.  I think we should keep the slightly weird key as is.

src/ArchiveAdmin.php
emteknetnz: I'm not sure, I presume at either before or after the 3 special tabs.  I'm assuming that they add tabs via an extension and yml similar to https://github.com/silverstripe/silverstripe-versioned-admin/blob/1/_config/archive-admin.yml.  Probably after these 3 tabs though before 'Other'?  I would expect the uasort() function to not impact the user defined tabs.

bergice (approved):

PR #167, opened 2020-08-26
FIX: Allow use of ArchiveAdmin on unstaged objects.

sminnee:


NightJar (approved):
Nice work :)

tests/ArchiveAdminTest/SingleStageObject.php
```diff
@@ -0,0 +1,16 @@
...
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class SingleStageObject extends DataObject implements TestOnly
+{
+    private static $table_name = 'Test_SingleStageObject';
+
+    private static $extensions = [
```
emteknetnz, line 13: The used of `.versioned` may be a bit confusing for people (myself included) - might we worth putting an inline comment and/or link to https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/00_Model/10_Versioning.md#applying-the-versioned-extension-to-your-dataobject to say that in enables stageless versioning (_Versions table but not _Live table)

emteknetnz (changes requested):
Minor recommendation for inline comment

chillu:
Merging. The inline comment is not a blocker.

PR #174, opened 2020-09-29
NEW Extension to exclude ChangeSetItem from used on table

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1098

Blocked by https://github.com/silverstripe/silverstripe-admin/pull/1139


maxime-rainville (approved):
Wouldn't hurt to have a unit test to cover the extension, but that's good enough.

PR #175, opened 2020-10-01
Improvement: Update Versioned Admin menu title to "Archive". Fixes #156

### LINKED ISSUE #156: Model admin name should be 'Archive' not 'Archives'
I think this was a mistake but it should technically be 'Archive' not 'Archives'.

[User help on how we talk about the Archive](https://userhelp.silverstripe.org/en/4/creating_pages_and_content/archive/#archive)

Relates to https://github.com/silverstripe/silverstripe-cms/issues/1575

chrispenny:
Fixes #156

> This was a design mishap where it was updated in one place and not another causing inconsistent across the board when it was developed.

Also requires: https://github.com/silverstripe/silverstripe-versioned/pull/302

### DIFF
--- a/lang/de.yml
+++ b/lang/de.yml
@@ -5,8 +5,12 @@ de:
     COLUMN_ORIGIN: Ursprung
     COLUMN_TITLE: Titel
     COLUMN_TYPE: Typ
+    MENUTITLE: Archiv
     SELECT_EMPTY: Auswählen...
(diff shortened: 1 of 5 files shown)


dhensby (approved):

chrispenny:
PR updated based on the comment from Robbie requesting that we move these translation between the two modules:
https://github.com/silverstripe/silverstripe-versioned/pull/302#pullrequestreview-500047700

PR #182, opened 2020-10-23
BUG Fix issue with composer warning about PSR-4 paths

tractorcow:
Fixes a super minor issue

> Deprecation Notice: Class SilverStripe\VersionedAdmin\Tests\Controller\HistoryControllerFactory\HistoryControllerFactoryExtension located in ./vendor/silverstripe/versioned-admin/tests/Controllers/HistoryControllerFactoryTest/HistoryControllerFactoryExtension.php does not comply with psr-4 autoloading standard. It will not autoload anymore in Composer v2.0. in phar:///opt/bitnami/php/bin/composer.phar/src/Composer/Autoload/ClassMapGenerator.php:201

### DIFF
--- a/tests/Controllers/HistoryControllerFactoryTest.php
+++ b/tests/Controllers/HistoryControllerFactoryTest.php
@@ -1,7 +1,7 @@
 <?php
 namespace SilverStripe\VersionedAdmin\Tests\Controllers;
 
-use SilverStripe\VersionedAdmin\Tests\Controller\HistoryControllerFactory\HistoryControllerFactoryExtension;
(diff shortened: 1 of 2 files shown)


dhensby:
So this is technically a breaking change, right? But it's in tests so I don't think we need to be so stringent there.

It's also a pain. We *could* add a class in the old namespace with a deprecation notice in the constructor so that there was no breaking impact (and then remove it from master).

robbieaverill (approved):
I have no semver qualms with tests, they're excluded from our distributables anyway

tractorcow:
@robbieaverill can this be merged? Just checking all our outstanding blockers to moving to composer 2 :)

robbieaverill:
cc @Cheddam @emteknetnz

ScopeyNZ (approved):

PR #190, opened 2020-11-17
BUG don't disable bundle.js / bundle.css with no-cms install

### LINKED ISSUE #183: "Select a content type" seems to be inoperative.
Context: I'm using versioned-admin in a non-cms project (no pages / sitetree).

In the versioned admin, I see only one tab with "Others" and a gridfield showing an archive of the first selected class.

However, the dropdown to select another model seems to do nothing.

image

Should I expect the gridfield to refresh with the archive for the selected class?

tractorcow:
If installed in a framework-only installation, the missing JS and CSS causes the interface to break. This fix ensures that the dependencies are always loaded.

Fixes #183

### DIFF
--- a/_config/cms.yml
+++ b/_config/cms.yml
@@ -11,8 +11,3 @@ SilverStripe\CMS\Controllers\CMSMain:
   extensions:
     - SilverStripe\VersionedAdmin\Extensions\CMSMainExtension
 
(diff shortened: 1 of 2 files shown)


emteknetnz (approved):
LGTM

Travis failure is unrelated

PR #197, opened 2021-02-09
FIX Avoid attempting to diff object-based field values

Cheddam:
In particular, this stops TreeMultiselectField values from breaking the comparison UI.

A more in-depth fix would introduce support for diffing these values, but there are multiple issues to address before this can be implemented, such as DiffField not honoring custom setValue methods on FormField sub-classes.

Fixes silverstripe/silverstripe-framework#9843.
Fixes silverstripe/silverstripe-framework#9839.


emteknetnz:
@Cheddam plenty of red on travis behat test

Should target this at 1.7 since there's no new API surface

emteknetnz:
Have merged up

Issue #199, opened 2021-03-02
/tests/ should be export-ignore in .gitattributes


tractorcow:
`composer install --prefer-dist` is installing tests for this module. We probably should not be packaging them.

# Pull request
- [ ] https://github.com/silverstripe/silverstripe-versioned-admin/pull/200

maxime-rainville:
Looks like it was just a typo in `.gitattributes`

This should take care of it https://github.com/silverstripe/silverstripe-versioned-admin/pull/200

PR #200, opened 2021-03-08
MNT Don't include tests in distribution code

maxime-rainville:
Fixes https://github.com/silverstripe/silverstripe-versioned-admin/issues/199

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-/tests/
+/tests/ export-ignore
 /.tx export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-/tests/ export-ignore
+/tests export-ignore
 /.tx export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,4 +1,4 @@
-/tests/
+/tests/ export-ignore
```
emteknetnz, line 1: Note sure if it matters, though I think we may want this without the trailing slash similar to what admin has.  Guaranteed that it'll work without the trailing slash

https://github.com/silverstripe/silverstripe-admin/blob/1/.gitattributes#L1

emteknetnz (changes requested):
Minor recommended change

emteknetnz:
Travis failure is behat on graphql4 job only https://github.com/silverstripe/silverstripe-versioned-admin/pull/200/checks?check_run_id=2053303830

.gitattributes
maxime-rainville:
```suggestion
/tests export-ignore
```

emteknetnz (approved):

PR #217, opened 2021-06-29
Fix missing dutch language title for ArchiveAdmin

RVXD:
In the CMS the archives admin does not have a title in some languages. In my case NL dutch.
Note: since last update other languages have this issue too. Perhaps someone can add the other language translations.
Commit causing issue:
https://github.com/silverstripe/silverstripe-versioned-admin/commit/c320a788a6b88d1d655eb7555b31a328922b6226

### DIFF
--- a/lang/nl.yml
+++ b/lang/nl.yml
@@ -5,12 +5,12 @@ nl:
     COLUMN_ORIGIN: Oorsprong
     COLUMN_TITLE: Titel
     COLUMN_TYPE: Type
-    MENUTITLE: ''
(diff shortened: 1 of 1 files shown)


J-Pielage:
Can this be merged?

emteknetnz:
Translation files for versioned-admin are managed via https://www.transifex.com/silverstripe/silverstripe-versioned-admin/dashboard/ rather than github and are usually updated as part of our regular release process.  There are some translations in transifex for Dutch which need to be pulled in, I'll see about getting this done.

Issue #212, opened 2021-05-20
Compare diff isn't working


emteknetnz:
CWP-2.8.0-rc1

I can "compare" versions, though there is not a content diff i.e. the green/red diff of text changes between versions

Instead it seemed like it only showed the most recent version

I see this on `/admin/pages/history/show/6132`

![image](https://user-images.githubusercontent.com/4809037/118909777-9b23e600-b977-11eb-8c18-88e75074bf80.png)

The content 'diff text' is on the latest version, but not the older version that I'm comparing against

Tested in IE11 and Firefox

## Acceptance Criteria
- [x] Text added without a matching deleted sample is highlighted in green
- [x] Deleted text style is retained

## Notes
- This never worked in Safari before ... we don't expect it to.
- There's a weird code snippet that's been copied in our codebase to handle the diff generation. If it's possible to patch it, great! Otherwise we might need to look for an alternative library.

## Pull Requests
- #237

maxime-rainville:
I'm only getting a diff when content is deleted/replace.

When brand new content is added that doesn't get highlighted. 

# Initial version
![image](https://user-images.githubusercontent.com/1168676/119062613-5bb6d180-ba2b-11eb-88a6-1a4ad52f2ea6.png)

# Updated version
![image](https://user-images.githubusercontent.com/1168676/119062589-4772d480-ba2b-11eb-9242-79aff2041f72.png)

# Diff
![image](https://user-images.githubusercontent.com/1168676/119062564-35913180-ba2b-11eb-9194-e365dd18cb79.png)

brynwhyman:
This was called out as a regression but I'm wondering if it indeed is... There are some related issues out there already:
* [Diff doesn't show an entirely new line as an addition](https://github.com/silverstripe/silverstripe-versioned-admin/issues/82)
* [No colour indicating changes between blocks on pages](https://github.com/silverstripe/silverstripe-versioned-admin/issues/111)

emteknetnz:
Replicated in cms 4.7.x-dev - so it's not a regression, it's an existing issue

GuySartorelli:
## What's happening
This is a styling issue.  It is caused by the `<ins>` tag surrounding the new `<p>` tag and its content. Deletions of entire paragraphs also don't get the appropriate `background-color` applied, though the deleted content does get strikethrough applied.
[...]

PR #237, opened 2022-04-10
FIX Highlight paragraph changes in diff.

GuySartorelli:
Changes which affect entire paragraphs (or any other block-level elements) were not being highlighted because the `ins` and `del` elements are treated as `display:inline;` by default.

## Parent issue
- #212


emteknetnz (approved):
Test locally, works great

PR #257, opened 2022-10-10
API Strongly-type action method signatures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10466

### DIFF
--- a/src/Controllers/HistoryViewerController.php
+++ b/src/Controllers/HistoryViewerController.php
@@ -72,10 +72,8 @@ class HistoryViewerController extends LeftAndMain
      *
      * WARNING: Experimental API.
      * @internal
-     * @param HTTPRequest $request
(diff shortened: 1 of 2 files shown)


src/Controllers/HistoryViewerController.php
```diff
@@ -75,7 +75,7 @@ public function getClientConfig()
      * @param HTTPRequest $request
      * @return HTTPResponse
```
ChloeSartorelli:
```suggestion
```

GuySartorelli (changes requested):

src/Controllers/HistoryViewerController.php

GuySartorelli (approved):
Not to be merged until the other PRs are ready to avoid breaking CI

PR #270, opened 2023-01-25
FIX Remove unused import

GuySartorelli:
This should have been done when the original build stack pass was done for installer modules.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


GuySartorelli:
js CI failure is to be fixed in https://github.com/silverstripe/silverstripe-admin/issues/1421

emteknetnz (approved):

PR #290, opened 2023-06-02
Fix undefined variable when $models is empty in ArchiveAdmin.php

alessandromarotta:
Fix undefined variable when $models is empty

### DIFF
--- a/src/ArchiveAdmin.php
+++ b/src/ArchiveAdmin.php
@@ -313,10 +313,10 @@ class ArchiveAdmin extends ModelAdmin
      */
     public function getManagedModels()
     {
-        $models = $this->getVersionedModels();
+        $archivedModels = $this->getVersionedModels();
 
         // Normalize models to have their model class in array key and all names as the value are uppercased
-        foreach ($models as $k => $v) {
+        foreach ($archivedModels as $k => $v) {
             $archivedModels[$v] = array('title' => ucfirst(singleton($v)->i18n_plural_name() ?? ''));
             unset($archivedModels[$k]);
         }


src/ArchiveAdmin.php
```diff
@@ -313,6 +313,7 @@ public function Breadcrumbs($unlinked = false)
      */
     public function getManagedModels()
     {
+        $archivedModels = [];
         $models = $this->getVersionedModels();
```
michalkleiner: Looking at the logic further down, I think this is what it was meant to be before.
```suggestion
        $archivedModels = $this->getVersionedModels();
```

michalkleiner (changes requested):

src/ArchiveAdmin.php
```diff
@@ -313,10 +313,10 @@ public function Breadcrumbs($unlinked = false)
...
     {
-        $models = $this->getVersionedModels();
+        $archivedModels = $this->getVersionedModels();
 
         // Normalize models to have their model class in array key and all names as the value are uppercased
-        foreach ($models as $k => $v) {
+        foreach ($archivedModels as $k => $v) {
             $archivedModels[$v] = array('title' => ucfirst(singleton($v)->i18n_plural_name() ?? ''));
             unset($archivedModels[$k]);
         }
```
ChloeSartorelli, line 322: We're now modifying the array we're looping over - could this introduce problems such as accidentally skipping items, or running over an item a second time (if `$archivedModels[$v]` gets added to the end of the array because it wasn't there before), or ending up with both a `$archivedModels[$k]` _and_ `$archivedModels[$v]` entry where before we only had `$archivedModels[$v]` in the result array?

GuySartorelli:
@alessandromarotta Are you still interested in having this PR merged? There are some questions that still need to be answered.

alessandromarotta:
Hello @GuySartorelli yes of course, I'm still interested in having this bug fixed.

There is a variable being invoked without being declared first. If $models is empty, $archivedModels will not be defined. It's quite evident.

image

src/ArchiveAdmin.php
ChloeSartorelli: @alessandromarotta Can you please address the concerns in my above comment?

GuySartorelli:
If literally the only problem is that `$archivedModels` is not defined, why don't we just add a `$archivedModels = [];` before the loop?

Edit: Nevermind, I see that was the original suggested solution.

GuySartorelli:
I'm going to close this for now since there has been no action for several months - we can reopen it when/if you get time to pick it up again.

Issue #291, opened 2023-06-04
Trailing slash expected for other model selector field


thats4shaw:
The other model selector field adds a trailing slash to the `data-others-archive-url` attribute however the framework now strips it out by default since [CMS5](https://docs.silverstripe.org/en/5/changelogs/5.0.0/#trailing-slash). 

This causes any selection of this dropdown to return a 404.

If trailing slashes are reenabled everything works again.  Will add a PR when I get a spare moment but wanted to document so I don't forget :-).

## Acceptance criteria
- You can select other models from the drop down in the "Archive" CMS section
- The scenario is covered by a Behat test

## Notes
- We are talking about the drop down on this page `admin/archive?others=1`

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/293

GuySartorelli:
PR merged and tagged as [2.0.1](https://github.com/silverstripe/silverstripe-versioned-admin/releases/tag/2.0.1)

PR #293, opened 2023-06-14
FIX Handle no trailing slash in other tab of archive admin

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/291

I've made some other changes to some [seemingly unrelated behat failures](https://github.com/silverstripe/silverstripe-versioned-admin/actions/runs/5262808189/jobs/9513413049?pr=293) that were happening only in the php 8.2 / mysql 8.0 job.  Seems like for whatever reason that "Home" page wasn't getting a page ID of 1, so I've updated the behat tests to not rely on the page ID to be 1. Normally I wouldn't do this in the same PR since it's unrelated, though even I wasn't sure if it was actually unrelated or not to the code change.


client/src/legacy/ArchiveAdmin/ArchiveAdmin.js
```diff
@@ -22,7 +22,7 @@ jQuery.entwine('ss', ($) => {
       const othersArchiveUrl = $(this).data('others-archive-url');
       if (targetClassName) {
         targetClassName = targetClassName.replace(/\\/g, '-');
-        const url = othersArchiveUrl + targetClassName;
+        const url = `${othersArchiveUrl}/${targetClassName}`;
```
ChloeSartorelli: This assumes `othersArchiveUrl` will now _never_ have a trailing slash.... use `joinUrlPaths()` instead per https://docs.silverstripe.org/en/5/changelogs/5.0.0/#trailing-slash

src/ArchiveAdmin.php
```diff
@@ -282,7 +282,7 @@ public function getOtherModelSelectorField($currentModel = '')
         );
         $modelSelectField->setAttribute(
             'data-others-archive-url',
-            $this->Link('/')
+            rtrim($this->Link('/'), '/')
```
ChloeSartorelli: Don't do this. Respect trailing slash config.

tests/Behat/features/archive-other.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Achive admin other tab
```
ChloeSartorelli:
```suggestion
Feature: Archive admin other tab
```
Super minor but could affect searching for "archive admin" across all files in the future so I figured I'd point it out. Won't hold back merging for it.

GuySartorelli (changes requested):

client/src/legacy/ArchiveAdmin/ArchiveAdmin.js
emteknetnz: Updated

src/ArchiveAdmin.php
emteknetnz: Updated

tests/Behat/features/archive-other.feature
emteknetnz: Updated

GuySartorelli (approved):
Looks good and works well locally.

Issue #300, opened 2023-08-03
"Files" tab in archive admin is missing


jenniferaslan:
The documentation shown here describes functionality that is not showing in a SilverStripe 4.13 site I recently built and launched: 

https://github.com/silverstripe/silverstripe-userhelp-content/blob/4/docs/en/02_Creating_pages_and_content/05_Archive.md#restoring-content-to-your-website

I have added the following to my app.yml file:
 ```
SilverStripe\Assets\File:
  keep_archived_assets: true
  show_history: true
```
However, after flushing, there is no Files tab in the Archive, nor is there any Archive option on files themselves in the Files interface. Is this functionality deprecated or broken?

## Acceptance criteria
- Validate that this was not removed intentionally.
- Make sure user help in in line with the capabilities of the CMS.
- If the capability is restored, make sure it is driven by the `keep_archived_assets` flag.

## Notes
- If you don't have the `keep_archived_assets` flag set to true, this feature will be less useful because you won't have any files to restore. Maybe it got intentionally remove because of that.

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/301
- https://github.com/silverstripe/developer-docs/pull/313
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1378
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1379

GuySartorelli:
Thanks for reporting this. I've moved this to the versioned-admin repository, as it seems like a genuine regression.

emteknetnz:
I've raised a [PR](https://github.com/silverstripe/silverstripe-versioned-admin/pull/301) to get the `SilverStripe\Assets\File.keep_archived_assets` config working properly as documented.

Note that `SilverStripe\AssetAdmin\Forms\FileFormFactory.show_history` is an asset-admin setting, the example in the original issue description incorrectly has this on `SilverStripe\Assets\File`

GuySartorelli:
@emteknetnz To satisfy the acceptance criteria "Make sure user help in in line with the capabilities of the CMS." the screenshot(s) on the linked userhelp page showing an "archive" button should be updated to show a "delete" button, and the docs should explain that "delete" means "archive" if old versions of files are retained.
Alternatively, the "delete" button should be replaced with an "archive" button when `keep_archived_assets` is true.

GuySartorelli:
This issue has now been resolved and will be automatically tagged once CI has finished running on the branch.

PR #301, opened 2023-08-07
ENH Allow File.keep_archived_assets to show files archive tab

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/300

This should have been checked `File.keep_archived_assets` all along, instead of AssetControlExtension

AssetControlExtension is an extension added to DataObject which is how the `keep_archived_assets` config is on File, so I can see why people though the config should have been read from there, though I think it was an error. We should that config working though for BC

Documentation has it on File - https://docs.silverstripe.org/en/4/developer_guides/files/file_management/#configuring-file-archiving

All the unit-tests that reference 'keep_archived_assets' use `File::config()->set('keep_archived_assets')`
- silverstripe/assets RedirectKeepArchiveFileControllerTest
- silverstripe/assets NormaliseAccessMigrationHelperWithKeepArchivedTest
- silverstripe/assets SS4KeepArchivedFileMigrationHelperTest

### DIFF
--- a/src/Extensions/FileArchiveExtension.php
+++ b/src/Extensions/FileArchiveExtension.php
@@ -77,6 +77,9 @@ class FileArchiveExtension extends DataExtension implements ArchiveViewProvider
     */
     public function isArchiveFieldEnabled()
     {
-        return Config::inst()->get(AssetControlExtension::class, 'keep_archived_assets');
+        // This should really only check File.keep_archived_assets, though it was originally
+        // only checking AssetControlExtension.keep_archived_assets, so keeping that for BC
+        return Config::inst()->get(AssetControlExtension::class, 'keep_archived_assets')
+          || Config::inst()->get(File::class, 'keep_archived_assets');
     }
 }
--- a/tests/Extensions/FileArchiveExtensionTest.php
+++ b/tests/Extensions/FileArchiveExtensionTest.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace SilverStripe\VersionedAdmin\Tests\Extensions;
+
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Assets\AssetControlExtension;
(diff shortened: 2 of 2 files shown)


src/Extensions/FileArchiveExtension.php
```diff
@@ -73,10 +73,14 @@ public function getArchiveField()
      * The files archive is only useful if archived assets are stored
      * so this checks if this option is enabled
      *
+     * This should really only check File.keep_archived_assets, though it was originally
+     * only checking AssetControlExtension.keep_archived_assets, so keeping that for BC
+     *
```
ChloeSartorelli:
```suggestion
```
This belongs as a code comment, not part of the PHPDoc

src/Extensions/FileArchiveExtension.php
```diff
@@ -73,10 +73,14 @@ public function getArchiveField()
...
      * @return boolean
     */
     public function isArchiveFieldEnabled()
     {
-        return Config::inst()->get(AssetControlExtension::class, 'keep_archived_assets');
+        return Config::inst()->get(AssetControlExtension::class, 'keep_archived_assets')
```
ChloeSartorelli, line 82:
```suggestion
        // This should really only check File.keep_archived_assets, though it was originally
        // only checking AssetControlExtension.keep_archived_assets, so keeping that for BC
        return Config::inst()->get(AssetControlExtension::class, 'keep_archived_assets')
```

GuySartorelli (changes requested):

src/Extensions/FileArchiveExtension.php
emteknetnz: Updated

GuySartorelli (approved):
Looks good, will merge when the remaining PRs are ready

PR #340, opened 2024-05-16
API Set extension hook implementation visibility to protected

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10514

Try re-running after other PRs are merged, so if that fixed failing unit test


GuySartorelli (approved):
LGTM

PR #347, opened 2024-06-17
ENH Allow no recordId to be passed in

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1782

While this isn't exactly the deepest way of fixing the issue, the component rendering is triggered by redux so it's going to be very difficult and not worth the effort just to fix a dev only console warning.


GuySartorelli (approved):
It seems like there's something very buggy about the `HistoryViewerField` - the entire `Root.History` tab is missing from inline-edible blocks, despite no code that I can find to exclude it.

Any other arbitrary tab you add via `getCMSFields()` is included, so without code to explicitly exclude this one, it should also display.

Adding a `HistoryViewerField` to another tab doesn't stop that tab from appearing in the list - but trying to open that field does cause the UI to freak out, and close the block. Trying a second time will open the tab, but without any history viewer at all.

All of that said - I don't think that's worth trying to tackle, especially not just to remove this one console error.
I had a click around the history tab of both non-inline-editable blocks and various pages and couldn't find any problem that this change causes, so I'm okay with it.

PR #353, opened 2024-07-16
MNT Split behat jobs

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/278


GuySartorelli (approved):
Jobs are around 7 and 5 mins respectively, so that's a fine split.

PR #354, opened 2024-07-26
API Remove GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799


src/Controllers/HistoryViewerController.php
```diff
@@ -56,16 +64,88 @@ class HistoryViewerController extends LeftAndMain
...
+            'revert' => $this->Link('api/revert'),
+            'read' => $this->Link('api/read'),
+        ];
         return $clientConfig;
     }
 
+    public function apiRead(HTTPRequest $request): HTTPResponse
+    {
+        $id = (int) $request->getVar('id');
+        $recordClass = $request->getVar('recordClass') ?? '';
```
ChloeSartorelli: Should be consistent with naming and either have
- `recordClass` with `recordId`, or 
- `class` with `id`

ChloeSartorelli: Applies to `apiRevert` as well.

src/Controllers/HistoryViewerController.php
```diff
@@ -339,4 +419,31 @@ protected function scaffoldForm($formName, array $context = [], array $extra = [
...
     }
+
+    private function getPostedJsonValue(HTTPRequest $request, string $key): mixed
+    {
+        $data = json_decode($request->getBody(), true);
+        if (!array_key_exists($key, $data)) {
+            $this->jsonError(400);
+        }
+        return $data[$key];
+    }
```
ChloeSartorelli: You added something very similar (if not identical) to this in `AssetAdmin` - is there any sense in centralising this logic somewhere? Perhaps even on the request itself?

src/Controllers/HistoryViewerController.php
```diff
@@ -56,16 +64,88 @@ class HistoryViewerController extends LeftAndMain
...
+        if (!$obj->canEdit()) {
+            $this->jsonError(403);
+        }
+        if (!$obj->getAtVersion($toVersion)) {
+            $this->jsonError(400);
+        }
+        /** @var Versioned|DataObject $record */
+        $record = Versioned::get_latest_version($recordClass, $id);
+        $record->rollbackRecursive($toVersion);
+        return $this->jsonSuccess(204);
```
ChloeSartorelli, line 158: Why 204 instead of 200?

src/Controllers/HistoryViewerController.php
```diff
@@ -56,16 +64,88 @@ class HistoryViewerController extends LeftAndMain
...
+        $this->checkJsonCsrfToken($request);
+        $id = (int) $this->getPostedJsonValue($request, 'id');
+        $recordClass = $this->getPostedJsonValue($request, 'recordClass');
+        $toVersion = (int) $this->getPostedJsonValue($request, 'toVersion');
+        $obj = $this->getDataObject($id, $recordClass, false);
+        if (!$obj->canEdit()) {
+            $this->jsonError(403);
+        }
+        if (!$obj->getAtVersion($toVersion)) {
+            $this->jsonError(400);
```
ChloeSartorelli, line 153: If this is not 404 because the data was in a post body instead of in the "resource" (aka URL), add a code comment about that so we don't just swap it to 404 later on.

src/Controllers/HistoryViewerController.php
```diff
@@ -339,4 +419,31 @@ protected function scaffoldForm($formName, array $context = [], array $extra = [
...
+        }
+        $obj = $recordClass::get()->byID($id);
+        if (!$obj) {
+            $this->jsonError($use404 ? 404 : 400);
+        }
+        if (!$obj->hasExtension(Versioned::class)) {
+            $this->jsonError(400);
+        }
+        return $obj;
+    }
```
ChloeSartorelli: This feels very similar to the `getFileByID()` you added to `AssetAdminOpen` - should it be generalised and put somewhere central like `LeftAndMain`? That would reduce duplicate code and mean if we want to update it later we just need to do so in one place.

GuySartorelli (changes requested):

src/Controllers/HistoryViewerController.php
emteknetnz: I would say class with id, however there's loads of existing JS that uses recordClass including JS components that aren't in this PR, so I think we should just keep as is

src/Controllers/HistoryViewerController.php
emteknetnz: It's a POST request so using 400

src/Controllers/HistoryViewerController.php
emteknetnz: 204 is for success with no content returned i.e. empty body in response

src/Controllers/HistoryViewerController.php
emteknetnz: It's such a small amount of code I think the centralisation probably just increases complexity rather than reduces it

src/Controllers/HistoryViewerController.php
emteknetnz: It is, though it's currently only used by one class right now, and it's a private method so could easily be made more generic later

It's also specific to doing JSON requests, and it allows polymorphism, which we don't allow on any other controllers, so it seems pretty unlikely to be used anywhere else

src/Controllers/HistoryViewerController.php
ChloeSartorelli: As discussed IRL - we should use `dataClass` for this endpoint.
The pre-existence of `recordClass` as the name of the variable in javascript is irrelevant.

src/Controllers/HistoryViewerController.php
ChloeSartorelli: Looks like you ended up moving it into `LeftAndMain` 👍

src/Controllers/HistoryViewerController.php
ChloeSartorelli: It would literally be used elsewhere, in `AssetAdminOpen` in place of `getFileByID()`. But like you say it's private so we can rip it out later when we look at these again and say "why is this almost identical logic in multiple places?"

src/Controllers/HistoryViewerController.php

GuySartorelli (approved):
LGTM, will merge when other PRs are happy

GuySartorelli (approved):
LGTM, will merge when other PRs are happy

GuySartorelli (approved):
re-approved most recent changes

PR #385, opened 2025-01-14
API Rename FormField Value to getFormattedValue

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11468

### DIFF
--- a/tests/Forms/DiffFieldTest.php
+++ b/tests/Forms/DiffFieldTest.php
@@ -22,7 +22,7 @@ class DiffFieldTest extends SapphireTest
         $diffField->setComparisonField($newField);
         $diffField->setValue('old');
 
-        $this->assertMatchesRegularExpression('/^<del>old<\/del> *<ins>new<\/ins>$/', $diffField->Value());
+        $this->assertMatchesRegularExpression('/^<del>old<\/del> *<ins>new<\/ins>$/', $diffField->getFormattedValue());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


tests/Forms/DiffFieldTest.php
```diff
@@ -45,59 +45,53 @@ public function testEscaping(
...
-                ReadonlyField::class,
-                'Something',
-                'Something <strong>bold</strong>',
-                'Something <del>&lt;strong&gt; bold &lt;/strong&gt;</del>',
-                'Non HTML field is escaped'
+            'readonly-add-bold' => [
+                'className' => ReadonlyField::class,
+                'oldValue' => 'Something',
+                'newValue' => 'Something <strong>bold</strong>',
+                'expected' => 'Something <ins>&lt;strong&gt; bold &lt;/strong&gt;</ins>',
```
ChloeSartorelli, line 70: Why did the expected value of this test case change? Haven't looked at the others, I assume something has gone awry.

GuySartorelli (changes requested):

tests/Forms/DiffFieldTest.php
emteknetnz: The test (not the code) was literally back to front where it was testing things as if the new version was the old version and vice versa.  Note that in the test itself (along with the dataProvider) I've reversed things. I fixed it as unrelated refactoring.

GuySartorelli (approved):
Approved under the assumption that:
1. The old test would still pass now
1. The new test would have passed before

PR #388, opened 2025-01-30
ENH Take advantage of new functionality in Versioned

GuySartorelli:
Avoids duplicated code to do what Versioned now lets you do.
Requires https://github.com/silverstripe/silverstripe-versioned/pull/489

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949


emteknetnz (changes requested):
As per https://github.com/silverstripe/silverstripe-versioned/pull/489#pullrequestreview-2625574268 just use setDataQueryParam()

emteknetnz (approved):

PR #395, opened 2025-04-28
ENH Use versions cache for ArchiveAdmin

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

Requires both https://github.com/silverstripe/silverstripe-versioned/pull/501 and https://github.com/silverstripe/silverstripe-framework/pull/11707

To see its effect, archive a couple of pages and then view them in ArchiveAdmin with https://github.com/emteknetnz/silverstripe-db-query-counter/ installed

`Count: 4 - Query: SELECT DISTINCT "SiteTree_Versions"."ClassName", "SiteTree_Versions"."LastEdited", "SiteTree_Versions"."Created", "SiteTree_Versions"."CanViewType", "SiteTree_Versions"."CanEditType", "SiteTree_Versions"."Version", "SiteTree_Versions"."Title", "SiteTree_Versions"."URLSegment", "SiteTree_Versions"."MenuTitle", "SiteTree_Versions"."Content", "SiteTree_Versions"."MetaDescription", "SiteTree_Versions"."ExtraMeta", "SiteTree_Versions"."ShowInMenus", "SiteTree_Versions"."ShowInSearch", "SiteTree_Versions"."Sort", "SiteTree_Versions"."HasBrokenFile", "SiteTree_Versions"."HasBrokenLink", "SiteTree_Versions"."ReportClass", "SiteTree_Versions"."ParentID", "SiteTree_Versions"."ID", CASE WHEN "SiteTree_Versions"."ClassName" IS NOT NULL THEN "SiteTree_Versions"."ClassName" ELSE 'SilverStripe\\CMS\\Model\\SiteTree' END AS "RecordClassName", "SiteTree_Versions"."RecordID", "SiteTree_Versions"."WasPublished", "SiteTree_Versions"."WasDeleted", "SiteTree_Versions"."WasDraft", "SiteTree_Versions"."AuthorID", "SiteTree_Versions"."PublisherID" FROM "SiteTree_Versions" WHERE ("SiteTree_Versions"."RecordID" = ?) ORDER BY "SiteTree_Versions"."LastEdited" DESC, "SiteTree_Versions"."Version" DESC`

Will be reduced down to

`Count: 1 - Query: SELECT DISTINCT "SiteTree_Versions"."ClassName", "SiteTree_Versions"."LastEdited", "SiteTree_Versions"."Created", "SiteTree_Versions"."CanViewType", "SiteTree_Versions"."CanEditType", "SiteTree_Versions"."Version", "SiteTree_Versions"."Title", "SiteTree_Versions"."URLSegment", "SiteTree_Versions"."MenuTitle", "SiteTree_Versions"."Content", "SiteTree_Versions"."MetaDescription", "SiteTree_Versions"."ExtraMeta", "SiteTree_Versions"."ShowInMenus", "SiteTree_Versions"."ShowInSearch", "SiteTree_Versions"."Sort", "SiteTree_Versions"."HasBrokenFile", "SiteTree_Versions"."HasBrokenLink", "SiteTree_Versions"."ReportClass", "SiteTree_Versions"."ParentID", "SiteTree_Versions"."ID", CASE WHEN "SiteTree_Versions"."ClassName" IS NOT NULL THEN "SiteTree_Versions"."ClassName" ELSE 'SilverStripe\\CMS\\Model\\SiteTree' END AS "RecordClassName", "SiteTree_Versions"."RecordID", "SiteTree_Versions"."WasPublished", "SiteTree_Versions"."WasDeleted", "SiteTree_Versions"."WasDraft", "SiteTree_Versions"."AuthorID", "SiteTree_Versions"."PublisherID" FROM "SiteTree_Versions" WHERE ("SiteTree_Versions"."RecordID" IN (?, ?)) ORDER BY "SiteTree_Versions"."LastEdited" DESC, "SiteTree_Versions"."Version" DESC`

Also

`Count: 2 - Query: SELECT "Version" FROM "SiteTree_Live" WHERE "ID" = ?`
and
`Count: 2 - Query: SELECT "Version" FROM "SiteTree" WHERE "ID" = ?`

Will reduce down to

Count: 1 - Query: SELECT "ID", "Version" FROM "SiteTree" WHERE "ID" IN (?, ?)
and
Count: 1 - Query: SELECT "ID", "Version" FROM "SiteTree_Live" WHERE "ID" IN (?, ?)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": "^8.3",
         "silverstripe/admin": "^3",
-        "silverstripe/framework": "^6",
+        "silverstripe/framework": "^6.1",
         "silverstripe/versioned": "^3.1"
     },
     "require-dev": {
--- a/_config/archive-admin.yml
+++ b/_config/archive-admin.yml
@@ -4,6 +4,9 @@ Name: archiveadmin
 SilverStripe\Versioned\VersionedGridFieldItemRequest:
   extensions:
     - SilverStripe\VersionedAdmin\Extensions\ArchiveRestoreAction
+SilverStripe\Forms\GridField\GridField:
+  extensions:
+    - SilverStripe\VersionedAdmin\Extensions\GridFieldExtension
 ---
 Name: pagesarchive
 Only:
--- a/src/Extensions/GridFieldExtension.php
+++ b/src/Extensions/GridFieldExtension.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace SilverStripe\VersionedAdmin\Extensions;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\ORM\DataList;
+use SilverStripe\VersionedAdmin\ArchiveAdmin;
+
+class GridFieldExtension extends Extension
+{
+    protected function onBeforeRenderHolder(GridField $gridField, array $properties): void
+    {
+        $controller = $gridField->getForm()?->getController();
+        if (!$controller || !is_a($controller, ArchiveAdmin::class)) {
+            return;
+        }
+        $list = $gridField->getManipulatedList();
+        if (!is_a($list, DataList::class)) {
+            return;
+        }
+        $list->prepopulateCaches();
+    }
+}


src/ArchiveAdmin.php
```diff
@@ -143,13 +143,17 @@ public static function createArchiveGridField($title, $class)
         // content is in a bad state.
         $list = Versioned::getRemovedFromDraft($class)->sort('LastEdited DESC');
 
+        /** @var DataObject $modelClass */
+        Versioned::prepopulateVersionsCache($class, $list->column('ID'));
```
emteknetnz: move to GridField

GuySartorelli (changes requested):
In my test with 548 archived pages with 1000 versions each this quickly hits memory limits, where it doesn't do that without this PR.

Taking a look, this is getting the data for all 548 pages even though the gridfield will only show 15 at a time. I assume the asset-admin change will have a similar problem.

Given this PR removes only 4 queries, I wonder if the risk of performance regression is worth it? Though the asset-admin one is likely more valuable because it batches from 2 queries _per file_ to 2 queries total.

If you do choose to proceed, this will need to only cache the data for the current page of records rather than all records in the underlying list.

emteknetnz:
Have updated to limit the DataList based on gridstate

Some testing - in ArchiveAdmin with 5 archived pages - running flush=1 then logging query count

"paginate 2" refers to adding the following code to ArchiveAdmin::createArchiveGridField() below the `$config` block

```php
$paginator = $config->getComponentByType(GridFieldPaginator::class);
$paginator->setItemsPerPage(2);
```

103 with cache + datalist limit + no paginate <<< with PR
100 with cache + datalist limit + paginate 2

103 with cache + no datalist limit + no paginate <<< Guy discovered memory issues
100 with cache + no datalist limit + paginate 2

119 with no cache + no datalist limit + no paginate <<< Original, before all of these PRs
104 with no cache + no datalist limit + paginate 2

GuySartorelli:
> Have updated to limit the DataList based on gridstate

It still doesn't take into account sort, or frankly _any_ list manipulation from e.g. a custom gridfield component.

composer.json
```diff
@@ -22,8 +22,7 @@
     "require": {
         "php": "^8.3",
         "silverstripe/admin": "^3",
-        "silverstripe/framework": "^6",
-        "silverstripe/versioned": "^3"
```
ChloeSartorelli: Why is the dependency on versioned being removed??

emteknetnz: Accident I think, added back in

emteknetnz:
Updated to use an extension so that we're caching the manipulated list

GuySartorelli:
Note also that https://github.com/silverstripe/silverstripe-versioned/pull/511 reduces the database queries for the archive admin - may be worth re-checking how many queries there are after that gets merged.

emteknetnz:
No longer needed based on https://github.com/silverstripe/silverstripe-versioned/pull/500#issuecomment-2910747928

PR #398, opened 2025-05-13
FIX Display an error when model history has errors

GuySartorelli:
This is only tangentially related to the issue at hand - without the associated versioned PR, the history view is stuck visually loading even though nothing is loading, and an error has occurred.

This PR displays a clear error instead of just looking like it's loading.

image

Note that the error list is only displayed in dev mode, to aid debugging. In test/live mode users will only see the first part telling them something went wrong and to try refreshing.

## Issue

- https://github.com/silverstripe/silverstripe-versioned/issues/505


emteknetnz (changes requested):
I don't think I like the concept behind this PR. I can't think of anywhere else in the CMS where we have a UX component intentionally behave differently based on the configured environment type. Have dev and live behave differently is something we tend to try and avoid.

Also assuming there is a bug, from content author point of view the UX isn't really that much of an improvement. Instead of a blank screen you get a red "something went wrong" ... but there's still nothing the user can do about it. In both cases something obviously isn't working so :shrug: This seems like it's intended as a dev feature though we can just look at the network responses for that like we do for everything else.

Given there a PR to fix the actual bug, presumably we're not expecting any issues at this point so there's not really actually a need for this anyway?

GuySartorelli:
We didn't expect this failure, but it failed. Displaying a message to the user when something fails is better than the current behaviour. We don't know what might fail in the future, but we shouldn't pretend our code is perfect - something is bound to fail eventually. When that happens, a friendly message is better than a loading screen.

This is just error handling. We should handle errors.

emteknetnz (changes requested):
I guess some error handling is better that none, though the solution isn't really consistent with the rest of the CMS.

Maybe just make a red toast and use the existing string `"Admin.UNKNOWN_ERROR": "An unknown error has occurred."`. A red toast would be the most standard UX to handle errors we have.

Remove the the dev/live code spilt, I really don't like the idea of putting debugging info here when we do it nowhere else in the CMS. Standard there is to just look at the the response in the network panel

GuySartorelli:
I don't see what harm the dev-only information is doing, and it surfaces information you otherwise have to dive into a GraphQL response for - if you even think to look there, given it's a 200 response.
But I will remove that part if it will help get the error handling merged in.

I also don't see what's wrong with replacing a section that's loading forever with an error message, which indicates to the user that this is in fact no longer loading.
I'd prefer to keep that instead of a toast, since even with a toast, a user may think the loading indicator is indicating that something is still loading. Usually where toasts are used, there isn't _also_ a loading indicator.
If you need a precedent for displaying an error on the page like this, the asset-admin and upload fields do that. I believe the import from CSV modal in GridField also handles errors this way.

> use the existing string `"Admin.UNKNOWN_ERROR": "An unknown error has occurred."`

The only place I could find that uses that is in https://github.com/silverstripe/silverstripe-admin/blob/e6aba5cfbf3c51ea8d10fb69f2f3de9ab0b8ac82/client/src/containers/FormBuilderLoader/FormBuilderLoader.js#L365-L396 which only uses it as a fallback if there is no error message to display. So if we want to be consistent with that usage, we should output the error message that gets returned from GraphQL (basically take what I was doing in dev mode and just always output all of that context). I'm happy to do that if you like. I've left the message as-is for now pending your decision.

emteknetnz:
> I don't see what harm the dev-only information is doing

Like I say, it's inconsistent with the rest of of the 'dev-experience' which is basically "check the network panel".  Also dev behaving differently from live is generally bad and should be avoided

> So if we want to be consistent with that usage, we should output the error message that gets returned from GraphQL 

I'd rather not since we're protecting against unknown errors at this stage. I have no idea what graphql will report as a potential error message in live.  I'm under the assumption that the graphql messages are aimed at developers, not at content editors.  That's why I think a generic "An unknown error occured" is better here, because at least we know what the user experience will be

> Usually where toasts are used, there isn't also a loading indicator.

We can just remove the loading indicator and trigger the toast?

I'm not super fussed if it's toast or random red panel, I am more concered about contents of the error message, though it just seems that red toast is more standard than random red panel (which technically should get ux team review as it's new design, whereas red toast we don't need to)

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -54,6 +55,27 @@ class HistoryViewer extends Component {
...
+      }
+    }
+    // eslint-disable-next-line no-restricted-syntax
+    for (const [key, val] of Object.entries(this.state)) {
+      if (nextState[key] !== val) {
+        shouldUpdate = true;
+      }
+    }
+    return shouldUpdate;
+  }
```
ChloeSartorelli: I'm not really pleased about this, but without it the component kept updating even though neither the props or state were being updated, and that resulted in displaying the toast twice.

I'm sure there's a simple react-y way to avoid that (e.g. `useEffect()` if this was a functional component) but I'm not sure how with a class component.

ChloeSartorelli: (updated since then to add type checks because in some cases one or both weren't an object (undefined I think) which results in an error.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -368,7 +394,13 @@ class HistoryViewer extends Component {
...
 
   render() {
-    const { loading, compare, currentVersion, recordId } = this.props;
+    const { graphQLErrors, loading, compare, currentVersion, recordId } = this.props;
+
+    // Display errors if any
+    if (graphQLErrors.length > 0) {
+      this.props.toastActions.error(i18n._t('Admin.UNKNOWN_ERROR', 'An unknown error has occurred.'));
+      return null;
+    }
```
emteknetnz:
```suggestion
    // Do not proceed if there are graphql error. A toast message will be shown in componentDidMount()
    if (graphQLErrors.length > 0) {
      return null;
    }
```

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -54,6 +55,31 @@ class HistoryViewer extends Component {
     }
   }
 
+  /**
```
emteknetnz:
```suggestion
  componentDidMount() {
    // Display a toast if there were any graphql errors
    const { graphQLErrors } = this.props;
    if (graphQLErrors.length > 0) {
      this.props.toastActions.error(i18n._t('Admin.UNKNOWN_ERROR', 'An unknown error has occurred.'));
    }
  }

  /**
```

emteknetnz (changes requested):
This will sort out the double toast

client/src/components/HistoryViewer/HistoryViewer.js
ChloeSartorelli: That doesn't work because on mount there are no errors - the errors come in the first update. This did point me in the right direction though, so thank you!

Update: I've chosen to do both this and a toast in `componentDidUpdate()`. The one in `componentDidUpdate()` is the one that triggers in the browser in my local testing - but if somehow there was an error in the initial props, we need `componentDidMount()` to catch that one.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -34,26 +35,33 @@ class HistoryViewer extends Component {
...
       versions.goToPage(currentPage);
     }
   }
 
+  componentDidMount() {
+    // Display a toast if there were any pre-existing graphql errors
+    const { graphQLErrors } = this.props;
+    if (graphQLErrors.length > 0) {
+      this.props.toastActions.error(i18n._t('Admin.UNKNOWN_ERROR', 'An unknown error has occurred.'));
+    }
```
emteknetnz:
```suggestion
    const { graphQLErrors, toastActions } = this.props;
    if (graphQLErrors.length > 0) {
      toastActions.error(i18n._t('Admin.UNKNOWN_ERROR', 'An unknown error has occurred.'));
    }
```

emteknetnz: Minor thing

emteknetnz (changes requested):

client/src/components/HistoryViewer/HistoryViewer.js
ChloeSartorelli: Done

GuySartorelli:
Also rebased

emteknetnz (approved):

PR #399, opened 2025-05-15
FIX Allow viewing archive history of archived records

GuySartorelli:
This was failing because `DataObject::get($class)` doesn't return archived records.
I've changed this to now use the same logic the graphql resolvers used (had to discover it by stepping through - there's unfortunately not a super clear way I can point at the code and show it, because it's not clear how GraphQL gets to the code it gets to)

## Issue

- https://github.com/silverstripe/silverstripe-versioned/issues/505


emteknetnz (changes requested):
Looks fine though will merge after the new behat test in 5.4 has been merged/merged-up/rebased first

GuySartorelli:
This has been rebased on top of the behat test commit

emteknetnz (approved):

Issue #402, opened 2025-05-16
BUG Order of "Date Archived" column is incorrect


christopherdarling:
### Module version(s) affected

versioned admin 2.4.0 / silverstripe recipe cms 5.4

### Description

Tested in the latest 5.4 release and the order of the records in the "Archive" section of the CMS is incorrect. The dates displayed are correct but the order that they appear in appears to be done of the first version of the record creation rather than the last version.

![Image](https://github.com/user-attachments/assets/a8afd041-e2ff-448e-8703-aa6d9dee8345)

As above but with "->Ago()" removed to display the raw dates.

![Image](https://github.com/user-attachments/assets/e7d44cd6-b591-4de9-8a34-ebdc8e3279bd)

### How to reproduce

Archive a number of pages and observe the order that they appear in in the GridField list.

### Possible Solution

Looks like the SQL query for the list is wrong.  I haven't yet found a fix.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/silverstripe-versioned-admin/pull/405

[**Kitchen sink CI run**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/15245660023) - note that the versioned-admin behat failure is [also happening on the `3` branch](https://github.com/silverstripe/silverstripe-versioned-admin/actions/runs/15242314339) and isn't related to this PR set.

- https://github.com/silverstripe/silverstripe-versioned/pull/511
- https://github.com/silverstripe/silverstripe-cms/pull/3083
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/406
- https://github.com/silverstripe/silverstripe-behat-extension/pull/304

GuySartorelli:
As Steve points out in his PR this is ultimately caused by filtering on `WasDeleted = 0` when using query mode `latest_versions`, which explicitly takes the archive version data _out of the list_ when we want to search. The date being displayed is actually fetched separately (which is also why that column can't be sorted by the user).

We need to add an alternative to `latest_versions` to get this working correctly, which means unfortunately this needs to be solved in a minor release rather than a patch, so it'll remain broken in CMS 5 but will be fixed from CMS 6.1 onward.

emteknetnz:
Linked PRs merged

PR #405, opened 2025-05-19
FIX Sort records by archive datetime

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/402

This doesn't work, will close PR though leave linked to the issue in case it's helpful for anyone else

The `all_versions` is the version mode that will actually get the everything from the _Versions table which is what we need, however it will return two rows for records that were unpublished and then archived even when filtering on WasDeleted = 1.

Unversioned records will be WasDeleted = 1, WasDraft = 0, WasPublished = 0, so simply filtering on WasDeleted = 1, WasDraft = 1 doesn't work

Also we need to fully use the ORM to get this to work and not just get the list, then get the unique RecordID's, and then do another query based of _Version.ID, as ArchiveAdminTest::testGridFieldList() will fail because that test has an existing DataList in a GridField dynamically update after archive a record in the list

I was surprised that I didn't get 'Versioned.mode' => 'archive' to work, it seems like it's closest to the query we need

We possible will need an entirely new Versioned method to get this to work so that the query is exactly as we need it

### DIFF
--- a/src/ArchiveAdmin.php
+++ b/src/ArchiveAdmin.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\VersionedAdmin;
 
+use InvalidArgumentException;
 use SilverStripe\Admin\ModelAdmin;
 use SilverStripe\Control\Controller;
 use SilverStripe\Core\ClassInfo;
@@ -24,6 +25,7 @@ use SilverStripe\Versioned\Versioned;
 use SilverStripe\Versioned\VersionedGridFieldState\VersionedGridFieldState;
 use SilverStripe\VersionedAdmin\Interfaces\ArchiveViewProvider;
 use SilverStripe\View\ArrayData;
+use SilverStripe\ORM\DataList;
 
 /**
  * Archive admin is a section of the CMS that displays archived records
@@ -130,6 +132,9 @@ class ArchiveAdmin extends ModelAdmin
      */
     public static function createArchiveGridField($title, $class)
     {
+        if (!is_a($class, DataObject::class, true)) {
+            throw new InvalidArgumentException("Class $class in not a DataObject");
+        }
         $config = GridFieldConfig_Base::create();
         $config->removeComponentsByType(VersionedGridFieldState::class);
         $config->removeComponentsByType(GridFieldFilterHeader::class);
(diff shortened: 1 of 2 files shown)


src/ArchiveAdmin.php
```diff
@@ -130,48 +132,24 @@ public function getEditForm($id = null, $fields = null)
      */
     public static function createArchiveGridField($title, $class)
     {
+        if (!is_a($class, DataObject::class, true)) {
```
emteknetnz, line 135: Unrelated safety check, since it's a static method anything could call this

src/ArchiveAdmin.php
```diff
@@ -130,48 +132,24 @@ public function getEditForm($id = null, $fields = null)
...
+        }
         $config = GridFieldConfig_Base::create();
         $config->removeComponentsByType(VersionedGridFieldState::class);
         $config->removeComponentsByType(GridFieldFilterHeader::class);
         $config->addComponent(new GridFieldDetailForm);
         $config->addComponent(new GridFieldViewButton);
         $config->addComponent(new GridFieldRestoreAction);
         $config->addComponent(new GridField_ActionMenu);
-
-        $singleton = singleton($class);
```
emteknetnz, line 141: All this code has been moved to a private method so that it can be unit tested in isolation

src/ArchiveAdmin.php
```diff
@@ -369,4 +347,37 @@ public function getManagedModelTabs()
...
+     * Gets a Datalist for use on the archive gridfield
+     */
+    private static function getListForGridField(string $dataClass): DataList
+    {
+        /** @var DataObject $dataClass */
+        $obj = DataObject::singleton($dataClass);
+        /** @var DataList $list */
+        $list = $obj->get();
+        $baseTable = $obj->baseTable();
+        $list = $list->setDataQueryParam('Versioned.mode', 'all_versions');
```
emteknetnz: This was previously `latest_versions` it is now `all_versions`. This was done because `SilverStripe\Versioned\Versioned::augmentSQLVersionedLatest()` which `latest_versions` uses will add `'WasDeleted' => 0` to the query which is something we need in order to sort by the actual archive datetime

src/ArchiveAdmin.php
```diff
@@ -369,4 +347,37 @@ public function getManagedModelTabs()
...
+            $liveTable = $baseTable . '_Live';
+            $list = $list->leftJoin(
+                $liveTable,
+                "\"{$baseTable}\".\"ID\" = \"{$liveTable}\".\"ID\""
+            );
+        }
+        $list = $list->where("\"{$draftTable}\".\"ID\" IS NULL");
+        $versionsTable = $baseTable . '_Versions';
+        $list = $list->where("\"{$versionsTable}\".\"WasDeleted\" = 1");
+        $list = $list->orderBy("\"{$versionsTable}\".\"ID\" DESC");
```
emteknetnz: These 3 lines are all new and replace `$list = $list->sort('LastEdited DESC');`

Issue #408, opened 2025-06-11
Error when loading history tab for a record which has a `GridField` on it


Jianbinzhu:
### Module version(s) affected

4.4.2

### Description

Uncaught LogicException when trying to access history tab of a page which has a `GridField` on it (in this case the gridfield name was `ChildPages`):

> Could not find a react component for field "ChildPages"

### How to reproduce

On a brand new CMS 5.4 install with `silverstripe/versioned-admin: 2.4.0`.

1. Create a new Blog Page (or any page which has a `GridField`)
2. Make some changes and publish the page
3. Go to history tab & click on one of the entries

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/silverstripe-versioned-admin/pull/409

> [!IMPORTANT]
> After merging, reassign to Guy to update the `@dataProvider` syntax after merge-up

GuySartorelli:
Can reproduce. I'll look into it. Thanks for raising this.
Note that there's no error actually displayed - you have to check the network tab. The only indication anything went wrong is there's no form fields displayed.

GuySartorelli:
Turns out this is a problem for any page with a gridfield. I'm shocked nobody has brought this up so far... migrating the issue to `versioned-admin`

emteknetnz:
Linked PR has been merged, it will be automatically tagged shortly

PR #409, opened 2025-06-12
FIX Don't choke in history view with GridField

GuySartorelli:
I know adding config isn't _usually_ done in a patch release, but:
1. There's no risk of adding config causing any new regressions, so it is safe
2. It gives developers a clean way to exclude their custom non-react-able fields from the history view.

Simply excluding GridFields was the behaviour before the exception was added in CMS 5.2, and seems like a sensible behaviour.

## Issue
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/408

### DIFF
--- a/src/Controllers/HistoryViewerController.php
+++ b/src/Controllers/HistoryViewerController.php
@@ -9,6 +9,7 @@ use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\Form;
+use SilverStripe\Forms\FormField;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBDatetime;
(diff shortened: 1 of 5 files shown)


emteknetnz (changes requested):
Shouldn't be adding new API in 5.4, this sort of enhancement should be targetting CMS 6.1 instead

Also the linked issue is only concerned with GridField, we don't actually need new configurable API here

GuySartorelli:
> this sort of enhancement should be targetting CMS 6.1 instead

This isn't an enhancement, it's a bug fix.

> Also the linked issue is only concerned with GridField

yes, but the same problem will occur with any form field that doesn't have a react implementation.

emteknetnz (changes requested):
Should instead filter out components that are [SCHEMA_DATA_TYPE_STRUCTURAL](https://github.com/silverstripe/silverstripe-framework/blob/6/src/Forms/FormField.php#L93)

In CMS 5 GridField isn't, so you'll need to hardcode for that.  In CMS 6 it is.

GuySartorelli:
Oh good point, we can just not render the same fields that would otherwise get an exception thrown. I opted to use the same condition used in `FromSchema::recursivelyValidateSchemaData()` which is where the exception would otherwise be thrown.

emteknetnz (approved):
Works well

PR #411, opened 2025-06-24
FIX Update expected row for behat test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/421

Fixes https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/15813413734/job/44638848749#step:12:165

"And I should see "Archived" in the record column in version 1"

You can see in the screen shot on the broken build artifacts that "Archived" is there, it's just that the row order is now reversed (and presumably now correct?)

### DIFF
--- a/tests/Behat/features/archive-history.feature
+++ b/tests/Behat/features/archive-history.feature
@@ -19,4 +19,4 @@ Feature: View archived history
     # Check the version history
     When I click on "History" in the header tabs
     Then I should see a list of versions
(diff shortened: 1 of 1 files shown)


GuySartorelli:
"presumably" is scary here. Can you please check if the order is actually correct?

emteknetnz:
It looks correct, this is the screenshot of the failing test where it was looking for "archived" in version 1

image

GuySartorelli (approved):
LGTM

PR #419, opened 2025-08-19
DEP Update JS dependencies

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/436

Have added resolutions for glob and minimatch. yarn/npm does not have a nice thing like composer conflict that accepts something dynamic, like "<9", instead you're forced to pin to an exact version, which is really annoying since otherwise yarn will use allow multiple versions.

Having a look through the generated yarn.lock file, it seems like they're both only used for devDependencies

### glob

glob 11 which is the latest version requires node 20, while we're stuck on node 18

@storybook/builder-webpack4@6.5.16 wants glob-promise "^3.4.0"
glob-promise@^3.4.0 wants "@types/glob" "*"

I think this is why we end up with a random requirement of glob 11, despite the fact that a bunch of other requirements only want glob ^7 or glob ^8

The differences between glob 7 and 8 appear to be mostly around it no longer allowing `\` as a path seperator on windows, instead forcing the use of `/`. This has no impact for us, so I selected `8.1.0`.

### minimatch
globby@^9.2.0 wants @types/glob@^7.1.1
@types/glob@^7.1.1 wants "@types/minimatch" "*"

I think this is why we end up with a random requirement of minimatch 10, despite the fact that a bunch of other requirements only want minimatch ^3 or minimatch ^5

The differences between minimatch 3 and 5 again appear to be mostly around it no longer allowing `\` as a path seperator on windows, instead forcing the use of `/`. This has no impact for us, so I selected `5.1.6`.


GuySartorelli (approved):
LGTM

PR #423, opened 2025-08-25
MNT Better debug info when behat test fails

GuySartorelli:
Rather than re-outputting what you already know you were looking for, this now outputs what is *actually* in that column so you know what went wrong.

Makes debugging failures like https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/17197777224/job/48784803532 easier.

> And I should see "Archived" in the record column in version 1
>      Record column contains Archived
>      Failed asserting that false is true.

## Issue

- https://github.com/tractorcow-farm/silverstripe-fluent/issues/1005


GuySartorelli:
CI failure will be fixed in https://github.com/silverstripe/silverstripe-versioned-admin/pull/424

emteknetnz (approved):

Issue #425, opened 2025-09-23
Cannot revert to archived version


emteknetnz:
### Module version(s) affected

5.4, 6.0

### Description

Clicking on an archive version has a button that says "Revert to this version". Clicking on the button hangs and does nothing.

### How to reproduce

1. Unpublished and archive a page
2. Go to page history
3. Click on the Archived record
4. Click "Revert to this version"

### Possible Solution

Probably just don't show the button for archived records.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/430

GuySartorelli:
PR merged. Will be automatically tagged by GitHub Actions

PR #430, opened 2025-10-28
FIX Disable revert button for archived versions

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/425

This will disable the "Revert to this version" button when viewing versioned record where the version was archived


client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -187,7 +189,7 @@ class HistoryViewerVersionDetail extends PureComponent {
           />
 
           <div className={classnames(formClasses)}>
-            <FormBuilderLoader
+            <FormBuilderLoaderComponent
```
emteknetnz, line 192: Need to be able to mock this otherwise we cannot use the new RTL test

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -128,6 +126,8 @@ class HistoryViewerVersionDetail extends PureComponent {
       return null;
     }
 
+    const forceDisabled = version.deleted;
```
ChloeSartorelli, line 129: We don't need a whole variable for this (can just use `forceDisabled={version.deleted}` below) - but I won't hold up merging just for that.

GuySartorelli (approved):
LGTM, works as expected

Issue #426, opened 2025-09-29
Accessibility: Can't access various elements using keyboard


GuySartorelli:
It's impossible to access various elements using the keyboard:

1. "compare two versions" checkbox
    <img width="260" height="133" alt="Image" src="https://github.com/user-attachments/assets/05093d6d-23fa-48a8-a25d-3db95ec8b453" />
1. "exit" button to exit compare mode
    <img width="402" height="76" alt="Image" src="https://github.com/user-attachments/assets/68cb7130-7c55-4519-a08d-c83ac7719208" />

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/433

GuySartorelli:
PR merged

PR #433, opened 2025-11-04
FIX Focus states for compare mode

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/426


GuySartorelli (approved):
LGTM, works as expected.

Issue #435, opened 2025-11-11
Convert versioned-admin to functional components


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/436

emteknetnz:
@GuySartorelli did you forgot to click submit review?

GuySartorelli:
Yup lol, sorry

GuySartorelli:
PR merged

PR #436, opened 2025-11-11
ENH Refactor versioned-admin from class to functional components

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/435


client/src/components/HistoryViewer/tests/HistoryViewerVersionDetail-test.js
```diff
@@ -27,6 +27,8 @@ function makeProps(obj = {}) {
     },
     compare: false,
     previewState: 'edit',
+    recordClass: 'MyRecordClass',
```
emteknetnz, line 30: A couple of required props on HistoryViewerVersionDetail - this prevents console warnings during yarn test

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -9,20 +9,27 @@ import { messageType } from 'types/messageType';
...
+const HistoryViewerVersionList = ({
+  compareModeAvailable = true,
+  extraClass = 'history-viewer__table',
+  messages = [],
+  showHeader = true,
+  versions = [],
+  FormAlertComponent,
+  HeadingComponent,
+  VersionComponent,
+  currentVersion,
```
emteknetnz, line 21: Was not in prop types that was used in isVersionActive()

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -7,92 +7,100 @@ import PropTypes from 'prop-types';
...
+  FormBuilderLoaderComponent = FormBuilderLoader,
+  recordId,
+  recordClass,
+  schemaUrl,
+  ToolbarComponent,
+  version,
+  compare = false,
+  previewState,
+  onAfterRevert,
+  compareModeAvailable,
```
emteknetnz, line 26: This was in this.props in renderDetails()

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -7,92 +7,100 @@ import PropTypes from 'prop-types';
...
-  /**
-   * Return whether or not we should be displaying the preview component
+   * Return whether or not we should be comparing two versions
    * @returns {boolean}
    */
-  isPreviewable() {
-    const { isPreviewable } = this.props;
-    return isPreviewable && !this.isCompareMode();
-  }
+  const isCompareMode = () => compare && compare.versionFrom && compare.versionTo;
```
emteknetnz, line 32: Moved above another function that call its

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -7,92 +7,100 @@ import PropTypes from 'prop-types';
...
   /**
-   * Return whether or not we should be comparing two versions
+   * Return whether or not we should be displaying the preview component
    * @returns {boolean}
    */
-  isCompareMode() {
-    const { compare } = this.props;
-    return compare && compare.versionFrom && compare.versionTo;
-  }
+  const isPreviewableCheck = () => isPreviewable && !isCompareMode();
```
emteknetnz, line 38: Renamed from isPreviewable() to isPreviewableCheck() as there is a isPreviewable var

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -15,119 +15,111 @@ import {
...
+  };
 
   /**
    * When clicking on a version, render the detail view for it via a Redux action dispatch
    * passed through via a closure prop (onSelect)
    */
-  handleClick() {
-    const { onSelect, version, isActive, compare } = this.props;
-
+  const handleClick = () => {
```
emteknetnz, line 64: Moved up because called by function below

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -29,207 +29,157 @@ import i18n from 'i18n';
...
+  isRevertable,
+  actions,
+  onSelect,
+  onSetPage,
+  onResize,
+  toastActions,
+  recordClass,
+  recordClassSingularName,
+  previewState,
+}) => {
```
emteknetnz, line 55: The following were defined in propTypes, but are not included here as they were not used in the old code:
- contextKey
- offset

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -29,207 +29,157 @@ import i18n from 'i18n';
...
+  recordId,
+  page,
+  isRevertable,
+  actions,
+  onSelect,
+  onSetPage,
+  onResize,
+  toastActions,
+  recordClass,
+  recordClassSingularName,
```
emteknetnz, line 53: This was added, in the old code was used in renderPagination()

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: I also removed graphQLErrors - it's redundant code as we don't use graphql, and the endpoint we do use `HistoryViewerController` does not include a 'graphQLErrors' key

This also led to me removing toastActions because that was only used when there were graphQL errors

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -29,207 +29,157 @@ import i18n from 'i18n';
...
-  }
+  };
+
+  /**
+   * List view is when either no current version is set, or only one of the two versions is
+   * set for compare mode
+   *
+   * @returns {boolean}
+   */
+  const isListView = () => {
```
emteknetnz, line 88: Moved up because called by function below

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -29,207 +29,159 @@ import i18n from 'i18n';
...
+  toastActions,
+  recordClass,
+  recordClassSingularName,
+  previewState,
+}) => {
+  const [versions, setVersions] = useState([]);
+  const [totalCount, setTotalCount] = useState(0);
+
+  // Use a ref to track previous error count without causing re-renders
+  const prevErrorCountRef = useRef(0);
```
ChloeSartorelli: This is new.

In all of these refactors _please_ point out when adding things that weren't there before, including what they're used for and why you needed to add them. It will make reviewing a easier and faster if I don't have to figure this stuff out for myself.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
...
-
-  render() {
-    const { graphQLErrors, loading, compare, currentVersion, recordId } = this.props;
+    return renderVersionList();
+  };
 
-    // Do not proceed if there are graphql error.
-    // A toast message will be shown in componentDidMount() or componentDidUpdate()
+  /**
+   * Display a toast if there were any pre-existing graphql errors
```
ChloeSartorelli:
```suggestion
   * Fetch version data on mount
   * Display a toast if there were any pre-existing graphql errors
```
The comment about displaying a toast used to be below `refreshVersionData` and is only for the funcitonality below `refreshVersionData`. If we're documenting this effect as a whole, we need to consider it as a whole.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
...
+    prevErrorCountRef.current = graphQLErrors.length;
+  }, []);
 
-    if (!recordId) {
-      return null;
+  /**
+   * Display a toast if there were any new graphql errors
+   */
+  useEffect(() => {
+    if (graphQLErrors.length > prevErrorCountRef.current) {
```
ChloeSartorelli: Why not just put `graphQLErrors.length` as the arg in the dependencies array?

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
...
-    if (loading) {
-      return <Loading />;
+  /**
+   * Refresh version data when the page changes
+   */
+  useEffect(() => {
+    if (versions.length === 0) {
+      return;
     }
+    refreshVersionData();
```
ChloeSartorelli:
```suggestion
    if (versions.length > 0) {
      refreshVersionData();
    }
```
This is probably cleaner but I won't hold up merging if you leave it as is since the result is the same and both are easily readable.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
...
-      return this.renderCompareMode();
+  /**
+   * Reset the selected version when unmounting HistoryViewer to prevent data leaking
+   * between instances
+   */
+  useEffect(() => () => {
+    if (typeof onSelect === 'function') {
+      onSelect(0);
     }
+  }, [onSelect]);
```
ChloeSartorelli: This used to be in `componentWillUnmount` - shouldn't it be returned from the `useEffect` that has no dependencies?

client/src/components/HistoryViewer/HistoryViewer.js
ChloeSartorelli: In this case I think it's not needed - see https://github.com/silverstripe/silverstripe-versioned-admin/pull/436#discussion_r2520108576

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -29,207 +29,159 @@ import i18n from 'i18n';
...
 
   /**
    * Handles setting the pagination page number
    *
-   * @param {number} page
+   * @param {number} pageNum
    */
-  handleSetPage(page) {
-    const { onSetPage } = this.props;
+  const handleSetPage = (pageNum) => {
```
ChloeSartorelli, line 145: `pageNum` renamed - looks like `page` is the name of a prop so this avoids a collision.

Friendly reminder to please point these out so I don't have to look around for the reason myself.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -257,48 +209,45 @@ class HistoryViewer extends Component {
...
-  renderPagination() {
-    const totalCount = this.state.totalCount;
-    const limit = this.props.limit;
-    if (this.state.versions.length === 0 || totalCount <= limit) {
+  const renderPagination = () => {
+    if (versions.length === 0 || totalCount <= limit) {
       return null;
     }
-    const props = {
+    const paginatorProps = {
```
ChloeSartorelli, line 233: Renamed but there's no global `prop` var for it to clash with - not clear why this was changed? Especially since it wasn't renamed in `renderVersionDetail`?

Leave as is - but I think for future PRs in this chain of refactors if there's changes like this which you haven't pointed out I'll stop reviewing and send it back with a "please point out all the changes" comment.

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
    *
    * @returns {HistoryViewerVersionList|null}
    */
-  renderComparisonSelectionList() {
-    const { compare: { versionFrom }, ListComponent } = this.props;
-
+  const renderComparisonSelectionList = () => {
+    const versionFrom = compare.versionFrom || null;
```
ChloeSartorelli: Why did you add `|| null` here?

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +257,121 @@ class HistoryViewer extends Component {
...
-    const {
-      isInGridField,
-      ListComponent,
-      CompareWarningComponent,
-      compare,
-      compare: { versionFrom: hasVersionFrom },
-    } = this.props;
-
+  const renderVersionList = () => {
+    const hasVersionFrom = compare.versionFrom || null;
```
ChloeSartorelli: Why did you add `|| null` here?

client/src/components/HistoryViewer/HistoryViewerHeading.js
```diff
@@ -1,73 +1,60 @@
...
-        dropdownOpen: shouldOpen
-      };
-    });
-  }
+  const toggle = () => {
+    // If we are opening the dropdown, force focus onto the checkbox.
+    const shouldOpen = !dropdownOpen;
+    if (shouldOpen && checkboxRef.current) {
+      // setTimeout() is used to wait for the DOM to update after the state change.
+      setTimeout(() => checkboxRef.current.focus(), 0);
```
ChloeSartorelli, line 23: I think the reason this was being done in a callback inside the setState function instead of outside it was so that this timeout isn't set until the set state starts to actually happen. 

If I understand correctly, the callback happens _during the next update phase_, which means the timeout will always be after the state change has affected changes inside the component (rerendering etc). But if you leave this outside the callback, the timeout may happen _during_ the next update phase instead of after it, causing a race condition.

client/src/components/HistoryViewer/HistoryViewerToolbar.js
```diff
@@ -9,28 +9,30 @@ import Config from 'lib/Config';
...
-      isReverting: false,
-    };
-  }
+const HistoryViewerToolbar = ({
+  FormActionComponent,
+  ViewModeComponent,
+  isLatestVersion = false,
+  isPreviewable = false,
+  isRevertable = false,
+  onAfterRevert = () => {},
```
ChloeSartorelli: `onAfterRevert` wasn't in default props before - is giving it a default function correct, or should it be marked as `required` in proptypes?

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -15,119 +15,110 @@ import {
...
 
   /**
    * Return a string of HTML class names for the row (list item) element
    *
    * @returns {string}
    */
-  getClassNames() {
-    const { extraClass, isActive, compare, compare: { compareFrom, compareTo } } = this.props;
+  const getClassNames = () => {
+    const { compareFrom, compareTo } = compare || {};
```
ChloeSartorelli: Why was ` || {}` added?

client/src/components/HistoryViewer/HistoryViewerVersion.js
```diff
@@ -15,119 +15,110 @@ import {
...
-  }
+  };
 
   /**
    * When closing the version, return back to the list view via Redux action dispatch
    */
-  handleClose() {
-    const { onSelect, version, compare, compare: { versionFrom } } = this.props;
+  const handleClose = () => {
+    const { versionFrom } = compare || {};
```
ChloeSartorelli: Why was ` || {}` added?

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
```diff
@@ -173,45 +151,56 @@ class HistoryViewerVersionDetail extends PureComponent {
...
-        {this.renderPreview()}
-      </div>
-    );
-  }
-}
+  /**
+   * When new props are received (from Redux dispatch events), check whether the preview
+   * state changes. If so, we want to add or remove the legacy CSS modifier for the CMS
+   * north toolbar based on whether the view mode is "split" (add) or anything else (remove)
+   */
```
ChloeSartorelli: This comment should be on the other `useEffect` - this one is for mounting, not updating.

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -9,20 +9,27 @@ import { messageType } from 'types/messageType';
 import { versionType } from 'types/versionType';
 import { compareType } from 'types/compareType';
 
-class HistoryViewerVersionList extends PureComponent {
+const HistoryViewerVersionList = ({
```
ChloeSartorelli, line 12: Arguably to convert from a `PureComponent` we should wrap the whole thing in `memo`, e.g

```js
const MyComponent = memo(({ myProp }) => {
    // ...
    return "some rendered stuff here";
});
```

Probably not worth doing though and I'm not 100% confident that it would be without problems.

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -42,28 +49,23 @@ class HistoryViewerVersionList extends PureComponent {
    * @param {Object} version
    * @returns {boolean}
    */
-  isVersionActive(version) {
-    const { currentVersion, compare, compare: { versionFrom, versionTo } } = this.props;
-
+  const isVersionActive = (version) => {
+    const { versionFrom, versionTo } = compare || {};
```
ChloeSartorelli: Why was `|| {}` added?

client/src/components/HistoryViewer/HistoryViewerVersionList.js
```diff
@@ -42,28 +49,23 @@ class HistoryViewerVersionList extends PureComponent {
...
     const isCurrent = currentVersion && currentVersion.version === version.version;
     const isCompareFrom = versionFrom && versionFrom.version === version.version;
     const isCompareTo = versionTo && versionTo.version === version.version;
-
-    return (!compare && isCurrent) || isCompareFrom || isCompareTo;
-  }
+    return ((!compare && isCurrent) || isCompareFrom || isCompareTo) || false;
```
ChloeSartorelli, line 58: Why was this condition changed?

GuySartorelli (changes requested):

client/src/components/HistoryViewer/HistoryViewerVersionList.js
emteknetnz: Won't do at this stage, we're not adding useCallback() etc during this refactor. If we want to add this sort of optimisation it would make sense to make that it's own separate refactor.

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewerHeading.js
emteknetnz: I think it's fine as is. You can test it locally by tabbing to the compare mode button and pressing space, you'll see the checkbox is reliably focused

Note I tried calling it without the setTimeout() and it did not focus the checkbox

The function is being passed to the toggle prop of the reactstrap component, so presumably there's something being done there ensure the toggle callback is being called at the correct time i.e. after the open dropdown has rendered

client/src/components/HistoryViewer/HistoryViewerToolbar.js
emteknetnz: I don't think it matters it's only used in one place (HistoryViewerVersionDetails) and that passes a prop in, so just leave as is

client/src/components/HistoryViewer/HistoryViewerVersion.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewerVersion.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewerVersionDetail.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewerVersionList.js
emteknetnz: Removed

client/src/components/HistoryViewer/HistoryViewerVersionList.js
emteknetnz: compare (compareType) can be false. 
HistoryViewerVersionList-test.js adds a lot of tests include some that test it with compare = false. Previously that state was not tested.

client/src/components/HistoryViewer/HistoryViewer.js
ChloeSartorelli: It's still not the length in the dependencies array? The old condition was specifically "if there are more errors" (i.e. the number of errors matters but not necessarily the content) and now it's just "if the errors have changed in some way" - I don't know what the intention was of the original condition so I don't know if this change satisfies that intention or not.

Why has the condition changed? How do we know if it satisfies the intention of the original condition?

client/src/components/HistoryViewer/HistoryViewer.js
```diff
@@ -308,97 +254,111 @@ class HistoryViewer extends Component {
...
+  };
 
-  render() {
-    const { graphQLErrors, loading, compare, currentVersion, recordId } = this.props;
+  /**
+   * Refresh version data on initial mount
+   */
+  useEffect(() => {
+    refreshVersionData();
+    // Reset selected version on unmount
```
ChloeSartorelli:
```suggestion
    // Reset selected version on unmount to prevent data leaking between instances
```
Adding back the context from the original comment

client/src/components/HistoryViewer/HistoryViewer.js
ChloeSartorelli: Looks like the "mount" `useEffect` no longer displays a toast? Why was that change made? I am begging you to tell me when you make changes that we haven't discussed and give the reasoning without me having to ask you about it.

client/src/components/HistoryViewer/HistoryViewerToolbar.js
ChloeSartorelli: If that passes a prop in and it's the only usage and it doesn't matter whether there's a default or not, then we shouldn't be adding a default prop. Please remove it.

The whole point of this is to do a like-for-like refactor. Yes it doesn't reeeaaallly matter on this one component but there's over a hundred components that need to be refactored. I'm asking you to remove this change here as a reminder to not make the change in the first place going forward. If the change isn't needed, it shouldn't be made, because that makes reviewing easier and reducing ping pong.

client/src/components/HistoryViewer/HistoryViewerVersionList.js
ChloeSartorelli: Again this was specifically about converting from a `PureComponent` - not about making additional/new optimisations. Feel free to leave as is, I just wanted to point out that it would be in scope for this refactor.

client/src/components/HistoryViewer/HistoryViewerVersionList.js
ChloeSartorelli: > compare (compareType)

What is `compareType`?

> HistoryViewerVersionList-test.js adds a lot of tests include some that test it with compare = false. Previously that state was not tested.

That still doesn't explain why it was changed - and why `|| false` is the appropriate part to add. Won't the `|| false` part only happen if `((!compare && isCurrent) || isCompareFrom || isCompareTo)` has already evaluated falsy, making it redundant?

GuySartorelli (changes requested):
Oops sorry, forgot to submit review

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewerToolbar.js
emteknetnz: Updated

client/src/components/HistoryViewer/HistoryViewerVersionList.js
emteknetnz: Will leave as is

client/src/components/HistoryViewer/HistoryViewerVersionList.js
emteknetnz: Sorry I explained badly. Basically under some conditions the old statement would return `undefined` instead of `false` - the function is supposed to return a bool `HistoryViewerVersionList marks current version as active when not in compare mode` expects `false`, so will fail with `undefined` is returned.
 
You can replicate this fails as follows:

```js
const currentVersion = { version: 2, draft: false, published: true };
const versions = [
  { version: 1, draft: false, published: true },
  currentVersion,
  { version: 3, draft: false, published: true },
];
const version = versions[0];
const compare = false;
//
const { versionFrom, versionTo } = compare;
const isCurrent = currentVersion && currentVersion.version === version.version;
const isCompareFrom = versionFrom && versionFrom.version === version.version;
const isCompareTo = versionTo && versionTo.version === version.version;
const retOld = (!compare && isCurrent) || isCompareFrom || isCompareTo;
const retNew = ((!compare && isCurrent) || isCompareFrom || isCompareTo) || false;
//
console.log('retOld', retOld) // undefined
console.log('retNew', retNew) // false

```

ChloeSartorelli: Can you please add a comment to the code to explain the use of `|| false` here for when we look at this in the future? Something like this will do:
`// Convert falsy values like undefined to a proper boolean false`

emteknetnz: Added

GuySartorelli (commented):
These two related comments are the only ones remaining. I'm probably just missing some context about them, the code's probably fine.

- https://github.com/silverstripe/silverstripe-versioned-admin/pull/436#discussion_r2524812884
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/436#discussion_r2524806198

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Changed to graphQLErrors.length

client/src/components/HistoryViewer/HistoryViewer.js
emteknetnz: Not required, if you put a console.log() statement in the replacement for 'componentDidUpdate' you'll see that it's triggered on mount

If the componentDidMount is not removed, then the condition would trigger twice.

GuySartorelli (approved):
LGTM

Issue #437, opened 2025-11-12
HistoryViewerField doesn't support models that use Versioning without stages


jonom:
### Module version(s) affected

3.1.1

### Description

<details><summary>Click to see original problem</summary>

See [related issue](https://github.com/silverstripe/silverstripe-versioned/issues/534). The docs note that [staging can be disabled](https://docs.silverstripe.org/en/6/developer_guides/model/versioning/#:~:text=staging%20can%20be%20disabled) to track changes without a draft stage, but this module doesn't seem to support that use case, and trying to use a HistoryViewerField results in an empty table as the request for records errors out.

If I change [this line](https://github.com/silverstripe/silverstripe-versioned-admin/blob/3/src/Controllers/HistoryViewerController.php#L123) to return true, the request to get records works fine and other versions are displayed, but they all have a LIVE label, and the UX breaks once you try to open a record for comparison.

### How to reproduce

```php
<?php

use SilverStripe\Forms\FieldList;
use SilverStripe\ORM\DataObject;
use SilverStripe\Versioned\Versioned;
use SilverStripe\VersionedAdmin\Forms\HistoryViewerField;

class ThingWithChangeHistory extends DataObject
{
    private static $db = [
        'Title' => 'Text',
    ];

    private static $extensions = [
        Versioned::class . '.versioned',
    ];

    public function getCMSFields(): FieldList
    {
        $fields = parent::getCMSFields();
        $fields->addFieldToTab(
            'Root.History',
            HistoryViewerField::create('HistoryViewer')
        );
        return $fields;
    }
}
```

</details>

if I select a record, the request for (I assume) the diff gives a 500 error and no content is visible:

<img width="745" height="388" alt="Image" src="https://github.com/user-attachments/assets/d7b4aca1-2c35-4e59-93e1-4f610c769212" />

Likewise if I try to compare:

<img width="739" height="392" alt="Image" src="https://github.com/user-attachments/assets/1d96353e-2f86-48d9-84a7-06df03444b71" />

The same error is triggered in both cases, but via `/admin/historyviewer/schema/compareForm/` in the latter case:

```
ERROR [Emergency]: Uncaught Error: Call to a member function remove() on null
IN GET /admin/historyviewer/schema/versionForm//3432?RecordVersion=6&RecordClass=[REDACTED]&RecordID=3432
Line 117 in /[REDACTED]/vendor/silverstripe/versioned-admin/src/Forms/DataObjectVersionFormFactory.php

[...]

PR #441, opened 2025-11-30
FIX Resolve isLiveVersion for non-staged objects

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/437

**Multiple commits, do not squash merge**

There is an unrelated fix in this PR to fix CSS for the history viewer where the table displays too wide, cutting on the start and end of the table from showing properly.


tests/Controllers/HistoryViewerControllerTest.php
```diff
@@ -568,6 +626,20 @@ private function getCsrfTokenheader(): array
         ];
     }
 
+    public static function provideFieldsRemovedFromForm(): array
```
emteknetnz, line 629: Unrelated refactoring - this dataProvider was in the wrong place

src/Controllers/HistoryViewerController.php
```diff
@@ -469,4 +469,18 @@ private function removeUnrenderableFields(?Form $form): void
             $form->Fields()->removeByName($toRemove);
         }
     }
+
+    /**
+     * Determine if the given record is the live version
+     * Will return false if the record if the record has the Versioned extension
```
michalkleiner:
```suggestion
     * Will return false if the record has the Versioned extension
```

GuySartorelli:
Can you please give that first commit a more meaningful commit message than `FIX History viewer css`? What about the CSS? What was the purpose of the commit?

src/Controllers/HistoryViewerController.php
```diff
@@ -469,4 +469,18 @@ private function removeUnrenderableFields(?Form $form): void
...
+     * but is not staged so that the 'Live' badge does not show, as there's no concept
+     * of draft/live in those cases
+     */
+    private function getIsLiveVersion(DataObject $record): bool
+    {
+        if (!$record->hasStages()) {
+            return false;
+        }
+        return (bool) $record->isLiveVersion();
+    }
```
ChloeSartorelli, line 485: Why are we doing this explicitly here? This feels like a change that should be made in `isLiveVersion()` directly.

Also I think the logic is reversed - if there are no stages, there is no _draft_ version, only a live version (or old versions - which this will also need to account for, another reason why `isLiveVersion()` is probably the correct place to put the logic)

GuySartorelli (changes requested):

emteknetnz:
Have moved CSS issues to [new issue](https://github.com/silverstripe/silverstripe-versioned-admin/issues/442)

src/Controllers/HistoryViewerController.php
emteknetnz: You're probably right. Will close PR

emteknetnz:
PR should be in versioned, not versioned-admin

Issue #442, opened 2025-12-01
History viewer table is cropped outside of `CMSMain`


emteknetnz:
When working on https://github.com/silverstripe/silverstripe-versioned-admin/issues/437 - noticed that the history view table is cropped - note uses code example from description of linked issue.

Note that this is in non-CMSMain sections, e.g. in a model admin's gridfield edit form. It displays correctly for pages in `CMSMain`.

<img width="1306" height="352" alt="Image" src="https://github.com/user-attachments/assets/46d54687-9fab-4e39-805c-4442c9e29a4f" />

## PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/443

PR #443, opened 2025-12-01
FIX Do not crop history viewer table

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/442


GuySartorelli (changes requested):
I assume this is for data objects that aren't in a `CMSMain`?
This has broken the `CMSMain` history styling. Notice the new weird space on the sides of the history list in the second screenshot.

**before this PR**
image

**after this PR**
image

emteknetnz:
>> I assume this is for data objects that aren't in a CMSMain?

Ah yup

Will just close this PR

Issue #414, opened 2025-08-06
Accessibility: Status badges don't have enough contrast.


GuySartorelli:
[WCAG 2.2 AA 1.4.3](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html)

The "Draft" and "Modified" status badges don't have enough contrast.

It's not clear from the report if these are the _only_ status badges that fail contrast, nor whether it's for just the regular badge or also the `badge--breadcrumbs` variants, so we should check them all.

## Acceptance criteria

- [x] All status flag styles meet accessibility contrast standards

## PRs

- https://github.com/silverstripe/silverstripe-admin/pull/2084
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/448

cassiezoest:
### Draft/modified badges

Update the text/border colour for these to #C73C00 to meet the contrast requirement.

This is optional, but personally I feel like the background colour is too light... if we wanted to darken it a touch and keep a compliant contrast ratio:

Background: #FFEDE0
Text/border: #C23A00

### Custom badges

Default colours are all good contrast wise.

emteknetnz:
Linked PRs have been merged, they will be released as part of CMS 6.2

PR #448, opened 2026-01-05
ENH Improve contrast for status badges

GuySartorelli:
Went with Cassie's second suggestion of darkening the background as well as increasing contrast.

## Issue

- https://github.com/silverstripe/silverstripe-versioned-admin/issues/414


emteknetnz (changes requested):
If a page is selected and highlighted in the site tree, and it has say a modified badge on it, now the background of the modified badge doesn't match the selected highlighted

You'll need to update the selected highlight to match the updated status badge background colour

GuySartorelli:
Good catch. Admin has some of this styling for some reason including duplicated colour references.
I've updated so that modified and draft in both versioned-admin and admin refer to the admin variables and updated the variables to the new colours.

Note there's now a PR for admin that this PR probably needs for JS CI to go green.

client/src/styles/legacy/_badges.scss
```diff
@@ -2,14 +2,12 @@
   .badge.status-#{$label}  {
     background-color: $bgColor;
     border-color: $textColor;
-  }
-  .badge.status-#{$label}  {
     color: $textColor;
   }
```
ChloeSartorelli, line 6: Unrelated removal of duplicate selector

emteknetnz (approved):

PR #467, opened 2026-03-24
MNT Fix behat test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/483

Fixes https://github.com/silverstripe/silverstripe-installer/actions/runs/23419329817/job/68433340792#step:12:429

Look like issue may have been datetime related because FeatureContext::getSpecificVersion() was looking for the string '1' in the entire row text to match the version number, rather than just the column, and that row text contains a date time, but instead it was sometimes matching the data e.g. 06/01/2026, or 10:23AM.

Have updated to just use the row number instead

### DIFF
--- a/tests/Behat/features/reset-state.feature
+++ b/tests/Behat/features/reset-state.feature
@@ -18,8 +18,7 @@ Feature: Reset state
 
   Scenario: Viewing a version gets reset
     Given I click on "History" in the header tabs
-    # the "first" version aka "version 1" is actually the most recent version - it'll be a version with the number "4"
-    Then I should see 4 in the version column in version 1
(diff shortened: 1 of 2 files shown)


tests/Behat/features/reset-state.feature
ChloeSartorelli, line 1: Also update other usages in this same scenario (e.g. line 30) - otherwise we're asserting about two separate things.

GuySartorelli (changes requested):

tests/Behat/features/reset-state.feature
emteknetnz: Done

GuySartorelli (approved):
LGTM

Issue #478, opened 2026-08-17
Archive admin preview shows the "page not found" page instead of the archived content


satrun77:
### Module version(s) affected

3.2.0

### Description

Previewing an archived page in the archive admin renders the site's "page not found" error page in
the preview panel instead of the archived content.

The archive preview state builds its URL from the record's own `LastEdited`:

```php
// src/Navigator/SilverStripeNavigatorItem_ArchiveLink.php
public function getLink()
{
    $link = $this->record->PreviewLink();
    return $link ? Controller::join_links($link, '?archiveDate=' . urlencode($this->record->LastEdited ?? '')) : '';
}
```

But the record the archive admin hands to the navigator is the *deleted* version, not a content
version: `ArchiveAdmin::createArchiveGridField()` lists `Versioned::getRemovedFromDraft($class)`,
which runs in `removed_from_draft` mode and therefore resolves to the version written by
`Versioned::createDeletedVersion()` during `doArchive()`. Its `LastEdited` is the moment of
archiving, correct for the "Date Archived" column, wrong as a date to read the record at.

On the frontend, that date resolves straight back to that deleted version, which no query can
return:

1. `ReadingMode::fromQueryString()` turns `?archiveDate=<archived at>` into `Archive.<archived at>.Stage`.
2. `Versioned::augmentSQLVersionedArchive()` selects `MAX("Version")` where
   `"LastEdited" <= <archived at>` and `"WasDraft" = 1` — that is the deleted version.
3. `Versioned::augmentSQLVersioned()` adds `"WasDeleted" = 0`, so the matched version is filtered
   out and no row is returned.
4. `ModelAsController` can't find the page, throws a 404, and the error page renders in the iframe.

So the preview asks for the record at the exact instant it stopped existing. Reading it one version
earlier, the last version that still had content, shows what was archived.

### How to reproduce

1. Install a fresh CMS with versioned admin (6.2, `silverstripe/versioned-admin` 3.2.0).
2. Create a page titled "Test page" and publish it.
3. Wait a moment (so the archive timestamp differs from the last save), then unpublish and archive
   that page from the page actions.
4. Go to **Archive** and use the **View** action on the archived page, e.g.
   `/admin/archive/SilverStripe-CMS-Model-SiteTree/EditForm/field/Pages/item/<id>/view`.
5. The preview panel shows the site's "page not found" page. Its iframe URL is
[...]

PR #479, opened 2026-08-17
FIX Preview archived records at their last content version

satrun77:
## Description

**Expected:** the archive admin's preview panel shows an archived page as it was when it was archived.

**Observed:** the panel renders the site's "page not found" error page.

## Manual testing steps

1. Create a page, publish it, then wait for a few seconds so the timestamps are different, and then unpublish and archive it.
2. Go to `/admin/archive` and use the **View** action on the archived page.
3. Before: the preview panel shows the "page not found" page, and the iframe URL carries
   `?archiveDate=<time you archived it>`. After: the panel shows the page content, and the URL
   carries `?archiveDate=<time you last saved it>`.
4. Regression check: non-archived records are untouched: open any published page in **Pages** and
   confirm the Draft/Published preview states still work.

## Issues

- #478 

## Pull request checklist

- [x] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ ] CI is green

### DIFF
--- a/src/Extensions/SiteTreeArchiveExtension.php
+++ b/src/Extensions/SiteTreeArchiveExtension.php
@@ -2,14 +2,20 @@
 
 namespace SilverStripe\VersionedAdmin\Extensions;
 
+use SilverStripe\CMS\Forms\SiteTreeURLSegmentField;
 use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\GridField\GridFieldDataColumns;
 use SilverStripe\Core\Extension;
 use SilverStripe\Forms\GridField\GridFieldFilterHeader;
 use SilverStripe\ORM\FieldType\DBDatetime;
 use SilverStripe\Security\Member;
+use SilverStripe\Versioned\Versioned;
 use SilverStripe\VersionedAdmin\ArchiveAdmin;
 use SilverStripe\VersionedAdmin\Interfaces\ArchiveViewProvider;
+use SilverStripe\VersionedAdmin\Navigator\SilverStripeNavigatorItem_ArchiveLink;
 
 /**
  * Adds a archive view for Pages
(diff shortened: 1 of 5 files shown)


tests/Behat/Context/FeatureContext.php
ChloeSartorelli: There's a lot of changes in this file, but it's not clear why they're being made. Can you please explain?

src/Extensions/SiteTreeArchiveExtension.php
```diff
@@ -78,4 +84,48 @@ public function isArchiveFieldEnabled()
     {
         return true;
     }
+
+    /**
+     * Point the URL segment field at the archived page.
+     */
```
ChloeSartorelli: Is this change relevant to the issue of the preview not showing the correct page? Or are you fixing something separate here?

tests/Behat/Context/FeatureContext.php
satrun77: There were some failed tests due to flaky Behat tests. Would you like me to revert them?

src/Extensions/SiteTreeArchiveExtension.php
satrun77: Yes, relevant to the issue.

tests/Behat/Context/FeatureContext.php
ChloeSartorelli: Not if they're needed to fix failed tests, but can you please explain what the changes are for specifically? e.g. what was failing?

src/Extensions/SiteTreeArchiveExtension.php
ChloeSartorelli: Can you please explain how this affects the preview?

tests/Behat/Context/FeatureContext.php
satrun77: You can see the failed Behat tests at https://github.com/silverstripe/silverstripe-versioned-admin/actions/runs/32088831461/job/95566886847

There were **3 race conditions** in the module helpers:

1. Unexpected alert open: `{Alert text : }`
   **Issue**: After a step fills an HTML field, the CMS form becomes dirty. The next navigation then triggers the browser's `beforeunload` dialog.
   **Fix**: Added an `@AfterStep` hook, registered first in `behat.yml`, which accepts any open alert and clears `window.onbeforeunload`.
2. `javascript error: window.jQuery is not a function`
   **Fix**: Guarded the expressions with `window.jQuery && ...` so they wait while the CMS finishes loading.
3. HTML field `Content` not found
   **Issue**: The framework's `stepIFillInTheHtmlFieldWith()` interacts directly with TinyMCE without waiting for it to initialise.
   **Fix**: Added a `@BeforeStep` hook that waits for every `textarea.htmleditor` to be registered and initialised before running any step that mentions an HTML field.

Additional cleanup:

* Waits now target **actual content rather than its container**, for example waiting for a version row instead of the `.table`.
* `getSpecificVersion()` now **retries** instead of returning `null`, which previously resulted in `Call to a member function find() on null`.
* Replaced the hard-coded **3-second waits** with a single **10-second `WAIT_TIMEOUT` constant**, as CI runners can be slow

src/Extensions/SiteTreeArchiveExtension.php
satrun77: The link in the URLSegment field links to the incorrect URL. It links to `?stage=Stage`, but an archived record isn't on the draft stage. It can also be `null` when the parent was archived too.

**Before:** `http://localhost/child-page?stage=Stage` → "page not found"
**After:** `http://localhost/parent-page/child-page?archiveDate=2026-09-18+02:44:27`

So an author clicking the URL in the edit form gets a 404, while the preview panel next to it shows the page fine. This points the link at the same archived state and date the preview uses.

`testUrlSegmentFieldMatchesTheArchivePreviewState()` asserts the two URLs are identical.

tests/Behat/Context/FeatureContext.php
ChloeSartorelli: > Unexpected alert open: `{Alert text : }`

The usual fix for this is to either put the `@unsavedchanges` tag on the scenario (see https://github.com/silverstripe/silverstripe-elemental/blob/750d4e27b8bfc744362307e7631364924d3451f5/tests/Behat/features/edit-block-element.feature#L54 for an example) or just make sure the changes are saved before ending the scenario. Please pick one of these instead. By adding a `@AfterStep` you're hiding potential problems down the line.

The rest of that sounds fine.

satrun77: Added `@unsavedchanges`

ChloeSartorelli (approved):
LGTM, thank you for this

